PR #39, opened 2017-05-18
MULE-12449: Fixed dependencies for mule-plugins to avoid including it…

gsfernandes:
…s transitive dependencies to application classloader


aiannucci:
Can this be merged?

PR #41, opened 2017-05-19
MULE-12402: until-successful retries fail after application restart d…

pablokraan:
…ue to inner flow still stopped


integration/src/test/java/org/mule/test/routing/SynchronizedFlowRefOnMuleContextStartTestCase.java
```diff
@@ -0,0 +1,122 @@
...
+import org.mule.tck.probe.Prober;
+import org.mule.test.AbstractIntegrationTestCase;
+
+import java.util.Map;
+
+import javax.xml.namespace.QName;
+
+import org.junit.Test;
+
+public class SynchronizedFlowRefOnMuleContextStartTestCase extends AbstractIntegrationTestCase {
```
elrodro83: add Allure annotations

PR #42, opened 2017-05-19
MULE-12256: add tests for redundant "/" in base path

alepulver:


http/src/test/resources/http-listener-path-leading-slashes-config.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:http="http://www.mulesoft.org/schema/mule/http"
+      xsi:schemaLocation="
+               http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+               http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd">
+
+    <http:listener-config name="listenerConfigWithSlash" basePath="/">
```
afelisatti: What about one with no basePath?

http/src/test/resources/http-listener-path-leading-slashes-config.xml
```diff
@@ -0,0 +1,40 @@
...
+        <http:listener-connection host="localhost" port="${port2}" />
+    </http:listener-config>
+
+    <http:listener-config name="listenerConfigWithSlashPrefix" basePath="/prefix">
+        <http:listener-connection host="localhost" port="${port3}"/>
+    </http:listener-config>
+
+    <flow name="testFlow1">
+        <http:listener path="${path}" config-ref="listenerConfigWithSlash"/>
+        <set-payload value="${path}"/>
```
afelisatti: There's an attributes properties that you can use to assert that the correct requests are received regarding what you'd expect considering the basePath/path relationship.

PR #81, opened 2017-06-01
MULE-12606: Remove org.apache.commons from exported packages

pablokraan:


integration/src/test/resources/org/mule/test/el/expression-language-config.xml
```diff
@@ -15,7 +15,7 @@
     <configuration>
         <expression-language autoResolveVariables="false">
             <import name="loc" class="java.util.Locale" />
-            <import class="org.mule.runtime.core.util.StringUtils" />
+            <import class="org.apache.commons.lang.StringUtils" />
```
elrodro83: don't you need commons-lang as a compile dependency in the test module for this to work?

pablokraan: The dependency is on the app/test classloader... probably related to the runner bug that does not exclude container's transitive dependencies from the app's classloader. I added the dependency as test to avoid having broken tests once that bug is fixed

PR #90, opened 2017-06-05
MULE-12673: fix cases when an exception occurs in an error handler

alepulver:
do not merge yet

### DIFF
--- a/integration/src/test/resources/org/mule/test/integration/exceptions/on-error-nested-exceptions-config.xml
+++ b/integration/src/test/resources/org/mule/test/integration/exceptions/on-error-nested-exceptions-config.xml
@@ -22,11 +22,6 @@
                 </on-error-continue>
             </error-handler>
         </try>
-        <error-handler>
-            <on-error-continue>
-                <append-string-transformer message=" and again"/>
-            </on-error-continue>
-        </error-handler>
     </flow>
 
     <flow name="propagatesToOuterWithoutExceptionAndFails">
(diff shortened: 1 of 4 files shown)


integration/src/test/resources/org/mule/test/integration/exceptions/on-error-nested-exceptions-config.xml
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:test="http://www.mulesoft.org/schema/mule/test"
+      xsi:schemaLocation="
+       http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+       http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd">
+
+    <flow name="propagatesToOuterWithoutExceptionAndSucceeds">
+        <try>
```
dfeist: Is double try needed to reproduce this? Is it worth testing both with and without this?

alepulver: Double try is needed to test if an exception in an error handler is catched in an outer scope. Testing cases where it should return an exception or event was optional I suppose, just to make sure that in addition to handling it continue can turn the case into success.

PR #100, opened 2017-06-07
MULE-12609: Every exported Mule package must contain api

pablokraan:


schedulers-tests/pom.xml
```diff
@@ -29,7 +29,7 @@
         <dependency>
             <groupId>org.mule.services</groupId>
             <artifactId>mule-service-scheduler</artifactId>
-            <version>${project.version}</version>
+            <version>1.0.0-SNAPSHOT</version>
```
elrodro83: use the property from the parent pom

PR #106, opened 2017-06-12
MULE-12255: Migrate testing namespaces to plugins

elrodro83:
* move unit tests of testing components back to its own module


integration/src/test/java/org/mule/test/IntegrationTestCaseRunnerConfig.java
```diff
@@ -9,7 +9,8 @@
 
 import org.mule.test.runner.ArtifactClassLoaderRunnerConfig;
 
-@ArtifactClassLoaderRunnerConfig(sharedRuntimeLibs = {"org.mule.tests:mule-tests-unit", "org.mule.tests:mule-tests-functional"})
+@ArtifactClassLoaderRunnerConfig(sharedRuntimeLibs = {"org.mule.tests:mule-tests-unit", "org.mule.tests:mule-tests-functional",
+    "org.mule.tests:mule-tests-component-plugin"})
```
pablokraan: This seems wrong to me. Why the app has to share code that is provided by the plugin?

PR #118, opened 2017-06-19
MULE-12853: Soap Service: Migrate to CXF 3

juandesi:


soap-connect/integration-tests/pom.xml
```diff
@@ -14,6 +14,7 @@
     <packaging>jar</packaging>
 
     <properties>
+        <cxfVersion>3.0.13</cxfVersion>
```
marianogonzalez: this is weird... why adding this but not adding anyone who uses it?

juandesi: Because it was using this one: https://github.com/mulesoft/mule/pull/5040/files#diff-600376dffeb79835ede4a0b285078036L159.

marianogonzalez: and shouldn't we changed that one? it does look like it doesn't make sense anymore.... All the components using CXF are now isolated and using different versions. Keeping that property is very error prone

PR #123, opened 2017-06-20
MULE-11689: Remove echo-component

elrodro83:


http/src/test/resources/http-listener-expect-header-streaming-auto-string-config.xml
```diff
@@ -12,7 +12,7 @@
 
     <flow name="defaultFlow">
         <http:listener config-ref="listenerConfig"  path="/" responseStreamingMode="AUTO"/>
-        <echo-component/>
+        <object-to-string-transformer/>
```
pablokraan: Same question as in the compatibility PR

elrodro83: the original echo was converting the payload to string calling the same logic this transformer does. Some tests depend on that behavior.

PR #125, opened 2017-06-21
MULE-12609: Every exported Mule package must contain api

pablokraan:
_ Moving classes from org.mule.runtime.core.execution to api/internal
_ Moving classes from org.mule.runtime.core.notification to api/internal
_ Moving more classes from org.mule.runtime.core.management.stats to api/internal
_ Moving more classes from org.mule.runtime.core.config to api/internal


http/src/test/java/org/mule/test/module/http/functional/listener/HttpListenerLifecycleTestCase.java
```diff
@@ -44,15 +45,15 @@ protected String getConfigFile() {
 
   @Test
   public void stopOneListenerDoesNotAffectAnother() throws Exception {
-    ExtensionMessageSource httpListener = (ExtensionMessageSource) ((Flow) getFlowConstruct("testPathFlow")).getSource();
+    ExtensionComponent httpListener = (ExtensionComponent) ((Flow) getFlowConstruct("testPathFlow")).getSource();
```
elrodro83: why this change?

pablokraan: This was the only place outside extension support modules referencing ExtensionMessageSource, but it is not required to reference it as the only used method exists on ExtensionComponent.
It could be debated why if ExtensionComponent is API then why ExtensionMessageSource is not... I prefer to expose the minimum number of classes, unless you disagree

PR #147, opened 2017-06-30
MULE-11701: Remove custom-source element

lucianoRM:


integration/src/test/resources/org/mule/test/integration/notifications/message-processor-notification-test-flow.xml
```diff
@@ -262,8 +262,8 @@
 
     <flow name="composite-source">
         <composite-source>
-            <custom-source class="org.mule.test.core.context.notification.processors.MessageProcessorNotificationTestCase$TestMessageSource"/>
-            <custom-source class="org.mule.test.core.context.notification.processors.MessageProcessorNotificationTestCase$TestMessageSource"/>
+            <test:custom-source class="org.mule.test.core.context.notification.processors.MessageProcessorNotificationTestCase$TestMessageSource"/>
```
elrodro83: can you put the logic from this source into `SkeletonSource` to avoid having this class?

integration/src/test/resources/org/mule/test/construct/flow.xml
```diff
@@ -33,8 +33,8 @@
 
     <flow name="flow2">
         <composite-source>
-            <custom-source class="org.mule.test.construct.FlowConfigurationFunctionalTestCase$TestSimpleMessageSource"/>
-            <custom-source class="org.mule.test.construct.FlowConfigurationFunctionalTestCase$TestSimpleMessageSource"/>
+            <test:custom-source class="org.mule.test.construct.FlowConfigurationFunctionalTestCase$TestSimpleMessageSource"/>
```
elrodro83: can you put the logic from this source into `SkeletonSource` to avoid having this class?

integration/src/test/resources/org/mule/test/components/flow-initial-state.xml
```diff
@@ -1,21 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <mule xmlns="http://www.mulesoft.org/schema/mule/core" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:test="http://www.mulesoft.org/schema/mule/test"
       xsi:schemaLocation="
-       http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd">
+       http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd">
     <flow name="defaultFlow">
-        <custom-source class="org.mule.test.components.FlowStateTestCase$TestMessageSource"/>
+        <test:custom-source class="org.mule.test.components.FlowStateTestCase$TestMessageSource"/>
```
elrodro83: can you put the logic from this source into `SkeletonSource` to avoid having this class?

integration/src/test/resources/synchronized-flowref-mule-context-start-config.xml
```diff
@@ -16,7 +16,7 @@
     </flow>
 
     <flow name="flow2">
-        <custom-source class="org.mule.test.routing.SynchronizedFlowRefOnMuleContextStartTestCase$UnblockProcessingSource"/>
+        <test:custom-source class="org.mule.test.routing.SynchronizedFlowRefOnMuleContextStartTestCase$UnblockProcessingSource"/>
```
elrodro83: can you put the logic from this source into `SkeletonSource` to avoid having this class?

lucianoRM: I don't think that the logic of this test class should be in SkeletonSource. What we could do is move the class to org.mule.functional.api.component but since it's only used here i don't see the benefits of doing that.

integration/src/test/java/org/mule/test/routing/SynchronizedFlowRefOnMuleContextStartTestCase.java
```diff
@@ -88,37 +81,12 @@ private void prePopulateObjectStore() throws ObjectStoreException {
     objectStore.store(testMuleEvent.getCorrelationId(), testMuleEvent);
   }
 
-  public static class UnblockProcessingSource implements MessageSource, Startable {
+  public static class UnblockProcessingSource extends SkeletonSource {
 
     @Override
     public void start() throws MuleException {
       waitMessageInProgress.release();
```
elrodro83: don't need this latch, checking for the started flag in skeleton source should be good

integration/src/test/resources/org/mule/test/construct/flow.xml
```diff
@@ -33,8 +33,8 @@
 
     <flow name="flow2">
         <composite-source>
-            <custom-source class="org.mule.test.construct.FlowConfigurationFunctionalTestCase$TestSimpleMessageSource"/>
-            <custom-source class="org.mule.test.construct.FlowConfigurationFunctionalTestCase$TestSimpleMessageSource"/>
+            <test:custom-source class="org.mule.functional.api.component.SkeletonSource"/>
```
elrodro83: replace with the skeleton-source element

integration/src/test/java/org/mule/test/routing/SynchronizedFlowRefOnMuleContextStartTestCase.java
```diff
@@ -82,50 +73,18 @@ private void prePopulateObjectStore() throws ObjectStoreException {
...
-
-    }
-  }
-
   public static class TestMessageProcessorCallback implements EventCallback {
 
     @Override
     public void eventReceived(Event event, Object component, MuleContext muleContext) throws Exception {
-      if (waitMessageInProgress.await(0, MILLISECONDS)) {
+      if (((SkeletonSource) ((AbstractPipeline) muleContext.getRegistry().lookupFlowConstruct(FLOW_UNDER_TESTING)).getSource())
```
lucianoRM: Could not assert here directly because i think this callback is executed in a different thread than the test call itself (waitsForStartedMuleContextBeforeAttemptingToSendMessageToEndpoint()). Since the test thread would have finished before the assert done here, it would've never failed.

PR #149, opened 2017-07-03
MULE-11686: Move component and it's dependencies to compatibility module

elrodro83:
First pass

* Migrate `component` usages.
* Move `component` tests to compatibility.


integration/src/test/resources/org/mule/test/construct/sub-flow.xml
```diff
@@ -109,9 +107,7 @@
         <append-string-transformer message="x"/>
         <append-string-transformer message="y"/>
         <append-string-transformer message="z"/>
-        <component>
-            <singleton-object class="org.mule.tck.core.lifecycle.LifecycleTrackerCheckComponent"/>
-        </component>
+        <custom-processor class="org.mule.tck.core.lifecycle.LifecycleTrackerCheckProcessor"/>
```
pablokraan: Should not use test processor?

schedulers-tests/src/test/java/org/mule/test/integration/SchedulerTestCase.java
```diff
@@ -7,28 +7,33 @@
...
-import org.mule.tck.probe.Probe;
-import org.mule.test.AbstractIntegrationTestCase;
 
 import java.util.ArrayList;
 import java.util.List;
 
 import org.junit.Test;
 
-public class SchedulerTestCase extends AbstractIntegrationTestCase {
+public class SchedulerTestCase extends MuleArtifactFunctionalTestCase {
```
pablokraan: AbstractIntegrationTestCase defines a different runner configuration than MuleArtifactFunctionalTestCase. Did you run all the module tests to be sure that this works?

integration/src/test/resources/org/mule/test/construct/sub-flow.xml
elrodro83: The callback on the `test:processor` doesn't allow to change the payload, and this processor does update the payload. I migrated the cases where the payload is not modified or it is not needed by the test afterwards to use `test:component`, but that approach doesn't work for cases like this.

schedulers-tests/src/test/java/org/mule/test/integration/SchedulerTestCase.java
elrodro83: Yes it works. I moved the scheduler tests to the scheduler test module, so i had to make this change to make the tests runners consistent.

PR #151, opened 2017-07-03
MULE-9682 - Define the list of objects that we will allow to inject i…

pablolagreca:
…n the domain objects

### DIFF
--- a/integration/src/test/java/org/mule/test/integration/lifecycle/DependencyInjectionTestCase.java
+++ b/integration/src/test/java/org/mule/test/integration/lifecycle/DependencyInjectionTestCase.java
@@ -11,6 +11,7 @@ import static org.hamcrest.core.IsNull.notNullValue;
 import static org.junit.Assert.assertThat;
 import static org.mule.test.allure.AllureConstants.LifecycleAndDependencyInjectionFeature.ArtifactObjectsDependencyInjectionStory.ARTIFACT_OBJECTS_DEPENDENCY_INJECTION_STORY;
 import static org.mule.test.allure.AllureConstants.LifecycleAndDependencyInjectionFeature.LIFECYCLE_AND_DEPENDENCY_INJECTION;
+import org.mule.runtime.api.exception.MuleException;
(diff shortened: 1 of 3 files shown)


integration/src/test/java/org/mule/test/integration/lifecycle/DependencyInjectionTestCase.java
```diff
@@ -0,0 +1,54 @@
...
+@Stories(ARTIFACT_OBJECTS_DEPENDENCY_INJECTION_STORY)
+public class DependencyInjectionTestCase extends AbstractIntegrationTestCase {
+
+  @Override
+  protected String getConfigFile() {
+    return "org/mule/test/integration/lifecycle/artifact-object-dependency-injection-config.xml";
+  }
+
+  @Test
+  public void validateInjectedObjects() {
```
marianogonzalez: you're free not to do this change if you want, but I think this would be a lot clearer if each of this assertions move to its own test method

integration/src/test/java/org/mule/test/integration/lifecycle/DependencyInjectionTestCase.java
```diff
@@ -0,0 +1,54 @@
...
+import ru.yandex.qatools.allure.annotations.Features;
+import ru.yandex.qatools.allure.annotations.Stories;
+
+@Features(LIFECYCLE_AND_DEPENDENCY_INJECTION)
+@Stories(ARTIFACT_OBJECTS_DEPENDENCY_INJECTION_STORY)
+public class DependencyInjectionTestCase extends AbstractIntegrationTestCase {
+
+  @Override
+  protected String getConfigFile() {
+    return "org/mule/test/integration/lifecycle/artifact-object-dependency-injection-config.xml";
```
marianogonzalez: don't really need this. You could just create the instance and inject it with `DependencyInjectionBean 
 bean = muleContext.getInjector().inject(new DependencyInjectionBean());`

PR #165, opened 2017-07-07
MULE-12056-FIX: Move combine-collections-transformer to compatibility…

onavarrette:
… module.

- Update pom.xml to use the compatibility module for those test who need it.


integration/pom.xml
```diff
@@ -97,6 +97,13 @@
         </dependency>
 
         <dependency>
+            <groupId>com.mulesoft.mule.modules</groupId>
```
pablokraan: I told you that a CE component cannot use a EE component. You can't use this dependency here

integration/src/test/resources/scatter-gather-test.xml
```diff
@@ -17,8 +17,6 @@
             <set-payload value="banana"/>
```
pablokraan: Which test is using scatter-gather-test.xml?

lucianoRM: Removed the config file. I could see that the class that used it was ScatterGatherRouterTestCase but it does not exist anymore

PR #168, opened 2017-07-07
MULE-11575 Migrate use of blocking API by transports to use non-block…

dfeist:
…ing API that supports new interceptors. Step 2.


integration/src/test/java/org/mule/test/core/context/notification/processors/MessageProcessorNotificationTestCase.java
```diff
@@ -335,9 +335,9 @@ public void collectionAggregator() throws Exception {
     specificationFactory = () -> new Node()
         .serial(pre()) // open Splitter, unpacks three messages
         .serial(prePost()) // 1st message on Logger
-        .serial(prePost()) // gets to Aggregator
+        .serial(pre()) // gets to Aggregator
```
elrodro83: the changes in the core functionality shouldn't change this behaviour.

dfeist: Discussed this with others and correct behaviour. If there are any issues then we can re-visit.

PR #178, opened 2017-07-11
MULE-13031: Move HTTP test from integrations to connector

afelisatti:


integration/pom.xml
```diff
@@ -121,6 +121,14 @@
         </dependency>
 
         <dependency>
+            <groupId>org.mule.connectors</groupId>
+            <artifactId>mule-http-connector</artifactId>
```
elrodro83: why is this dependency required here?

afelisatti: Because of the HTTP authentication test I moved in order to avoid a cyclic dependency to spring-security. Ideally all those spring-security related tests should go here IMO.

PR #200, opened 2017-07-19
MULE-13113: Migrate spring security support to spring module

elrodro83:


integration/pom.xml
```diff
@@ -384,5 +386,11 @@
             <artifactId>commons-lang3</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
pablolagreca: why do you need access to this jar ?

elrodro83: the tests use the spring-security namespace to define beans. Since those resources are now not exposed by the plugin (https://github.com/mulesoft/mule-spring-module/pull/4#discussion-diff-128374653R39), the application has to include it.

PR #202, opened 2017-07-20
MULE-13007: Every exported Mule package must contain api (part3)

pablokraan:


integration/src/test/resources/non-blocking-test-config.xml
```diff
@@ -65,7 +65,7 @@
 
     <flow name="security-filter">
         <test:assert-intercepting responseSameTask="false" responseExpression="#[mel:assertPayload()]"/>
-        <custom-processor class="org.mule.runtime.core.processor.SecurityFilterMessageProcessor"/>
+        <!--<custom-security-filter ref="customFilter"/>-->
```
elrodro83: should there be a TODO to uncomment this?

PR #203, opened 2017-07-20
MULE-12057: DataWeave should be used to encrypt payloads in idempotent-secure-hash-message-validator

lucianoRM:


integration/src/test/resources/core-namespace-routers-flow.xml
```diff
@@ -5,13 +5,31 @@
        http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd">
 
     <flow name="IdempotentSecureHashReceiverRouter">
-        <idempotent-secure-hash-message-validator messageDigestAlgorithm="SHA-128"/>
+        <idempotent-message-validator idExpression="%dw 2.0
+            output text/plain
+            import dw::Crypto
+            ---
+            Crypto::hashWith(payload,'SHA-256')"/>
```
elrodro83: the test was using SHA-128 before. Is that available with DW here?

integration/src/test/resources/org/mule/test/integration/notifications/message-processor-notification-test-flow.xml
```diff
@@ -183,7 +183,11 @@
     </flow>
 
     <flow name="idempotent-secure-hash-msg-filter">
-        <idempotent-secure-hash-message-validator idExpression="#[id]" objectStore="objectStore"/>
+        <idempotent-message-validator idExpression="%dw 2.0
+        output text/plain
+        import dw::Crypto
+        ---
+        Crypto::SHA1(variables.id)" objectStore="objectStore"/>
```
elrodro83: i think the previous version was referring to the id of the event, not a variable

integration/src/test/resources/core-namespace-routers-flow.xml
lucianoRM: No, SHA-128 is not available within DW Crypto.

integration/src/test/resources/org/mule/test/integration/notifications/message-processor-notification-test-flow.xml
lucianoRM: I think that there is not Id as an event attribute. It should've been a variable

integration/src/test/resources/core-namespace-routers-flow.xml
lucianoRM: Actually, SHA-128 doesn't even exist. The test was just checking that the string was properly saved.

integration/src/test/java/org/mule/test/config/spring/DslConstantsRoutersFlowTestCase.java
```diff
@@ -30,13 +36,14 @@ public String getConfigFile() {
...
 
+  @Rule
+  public ExpectedException expectedException = none();
+
   @Test
   public void testIdempotentSecureHashReceiverRouter() throws Exception {
     Processor router = lookupMessageProcessorFromFlow("IdempotentSecureHashReceiverRouter");
-    assertThat(router.getClass().getName(),
-               equalTo("org.mule.runtime.core.internal.routing.IdempotentSecureHashMessageValidator"));
+    assertThat(router.getClass().getName(), equalTo(IdempotentMessageValidator.class));
```
elrodro83: this may cause classloader issues

PR #209, opened 2017-07-24
Mule 13170: runExpectingException() should handle cases where exception is thrown from a source different than the expected one

lucianoRM:


integration/src/test/resources/org/mule/functional/junit4/flow-runner-config.xml
```diff
@@ -0,0 +1,22 @@
...
+       http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd">
+
+    <flow name="okFailFlow">
+        <test:processor throwException="true"/>
+    </flow>
+
+    <flow name="badFailFlow">
+        <test:processor throwException="true"/>
+        <error-handler>
+                <on-error-propagate>
```
elrodro83: identation size

PR #212, opened 2017-07-25
MULE-12202: Don't map variables as single bindings

afelisatti:


integration/src/test/resources/foreach-test.xml
```diff
@@ -263,10 +263,10 @@
     </flow>
 
     <flow name="splitJsonComplexValue">
-        <foreach collection="#[content.users]" batchSize="2">
+        <foreach collection="#[vars.content.users]" batchSize="2">
             <foreach>
                 <test:queue name="splitJsonComplexValueOutQueue" contentJavaType="java.lang.String">
-                    <test:content>#["$(payload.name) $(payload.lastname)"]</test:content>
+                    <test:content>#['$(payload.name) $(payload.lastname)']</test:content>
```
lucianoRM: Single quotes?

afelisatti: I'll change it back.

PR #213, opened 2017-07-25
MULE-11716: Move set, remove, copy for attachments, properties, session properties to compatibility module

lucianoRM:


integration/src/test/resources/foreach-test.xml
```diff
@@ -91,7 +89,7 @@
             <test:processor/>
             <set-variable variableName="totalMessages" value="#[vars.counter]" />
         </foreach>
-        <set-property propertyName="totalMessages" value="#[vars.totalMessages]"/>
+        <test:assert expression="#[totalMessages == 3]"/>
```
elrodro83: add vars

PR #226, opened 2017-08-03
MULE-13138: Decouple runtime and spring module from spring libraries

elrodro83:


integration/pom.xml
```diff
@@ -179,6 +179,18 @@
...
+        <dependency>
+            <groupId>commons-collections</groupId>
+            <artifactId>commons-collections</artifactId>
+            <scope>test</scope>
+        </dependency>
+        
+        <dependency>
+            <groupId>com.github.stephenc.eaio-uuid</groupId>
+            <artifactId>uuid</artifactId>
+            <version>3.4.0</version>
```
pablokraan: Is not version managed on mule's poms?

elrodro83: no, it's declared explicitely in mule/core,

PR #235, opened 2017-08-09
MULE-11717: Revamp parse-template to support DW expressions

lucianoRM:


integration/src/test/resources/org/mule/transformers/parse-template-config.xml
```diff
@@ -58,4 +58,21 @@
         </parse-template>
     </flow>
 
+    <flow name="with-target-value-no-target">
+        <parse-template targetValue="#[payload]">
```
pablolagreca: can you add the message binding to DW and add a test case here?

PR #251, opened 2017-08-16
MULE-10580: Operation reconnection should be decoupled from config re…

marianogonzalez:
…connection


pablolagreca:
Do we have test cases for all connectors and for scenarios of source disconnection, connection disconnection, etc?

marianogonzalez:
no

pablolagreca:
Add those test cases

PR #262, opened 2017-08-18
MULE-13217 Rework foreach to be non-blocking and decoupled from splitter

dfeist:


integration/src/test/java/org/mule/test/routing/ForeachTestCase.java
```diff
@@ -374,7 +345,6 @@ public void jsonUpdate() throws Exception {
     order.put("email", "ellen.mail.com");
     order.put("items", items);
     flowRunner("process-json-update").withPayload(singletonMap("order", order)).run();
-    FlowAssert.verify("process-json-update");
```
elrodro83: why remove the verifies?

dfeist: Run verifies by default.

elrodro83: true that

PR #294, opened 2017-09-04
MULE-13428 Update foreach scope to not support Map

dfeist:


integration/src/test/resources/foreach-test.xml
```diff
@@ -60,7 +60,7 @@
     </flow>
 
     <flow name="map-config">
-        <foreach>
+        <foreach collection="#[dw:dw::core::Objects::entrySet(payload)]">
```
afelisatti: Why do you still need the prefix after this changes?

dfeist: missing push

PR #333, opened 2017-09-21
MULE-13322: flaky fix

lucianoRM:


integration/src/test/java/org/mule/test/integration/exceptions/LogCheckTestCase.java
```diff
@@ -26,6 +28,11 @@ protected String getConfigFile() {
     return "org/mule/test/integration/exceptions/log-check-config.xml";
   }
 
+  @BeforeClass
+  public static void disableFlowStackLog() {
+    System.setProperty(MULE_FLOW_TRACE, Boolean.toString(false));
```
afelisatti: What if a later test requires this?

afelisatti: Seems like you should use a `SystemProperty` rule.

PR #376, opened 2017-10-17
MULE-13644: Extract ConnectorLevelMessageDispatchingTestCase which uses the http service inside the mule project and generates a circular dependency

anosenzo:


aiannucci:
MERGE, DON'T SQUASH

PR #401, opened 2017-10-31
MULE-13912 - flow-ref to non defined flow does not fail deploy

pablolagreca:


integration/src/test/java/org/mule/test/construct/FlowRefBadConfigTestCase.java
```diff
@@ -0,0 +1,36 @@
...
+@Feature(CORE_COMPONENTS)
+@Story(FLOW_REFERENCE)
+public class FlowRefBadConfigTestCase extends AbstractMuleTestCase {
+
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  @Test
+  public void flowRefToNonExistentFlow() throws Exception {
+    expectedException.expectMessage(
```
elrodro83: see comments in CE

PR #416, opened 2017-11-15
MULE-13456: Fix flaky test org.mule.test.core.context.notification.processors.ComponentLocationTestCase.flowWithAsync

lucianoRM:


lucianoRM:
Only thing left to do was to remove the @ignore in the test but was removed in another PR

PR #418, opened 2017-11-16
MULE-14095: Soap connect transport should not be bounded in mule soap support

juandesi:


soap-connect/integration-tests/src/test/java/org/mule/it/soap/connect/InvokeSoapCustomTransportTestCase.java
```diff
@@ -59,15 +59,4 @@ public void sendThroughCustomTransportWithParams() throws Exception {
     String response = "<con:ram xmlns:con=\"http://ram.test.mule.org\"><text>CUSTOM RESPONSE</text></con:ram>";
     assertSimilarXml(response, (String) message.getPayload().getValue());
   }
-
-  @Test
-  public void withDefaultHttp() throws Exception {
```
marianogonzalez: why did you deleted this case? I would expect a new test to exist using your new provider

PR #422, opened 2017-11-20
MULE-14047 JMH Flow performance tests should use real scheduler service

dfeist:


performance/src/main/java/org/mule/test/FlowMixedAProcessorBenchmark.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.mule.runtime.core.api.processor.Processor;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
```
elrodro83: static imports

PR #453, opened 2017-12-28
MULE-14110: Add chainless component to ErrorHandler Router

alegmarra:


dsl/src/test/resources/core-bulk-extension-model.xml
```diff
@@ -47,16 +47,14 @@
...
-            <logger/>
-        </on-error>
+    <error-handler name="globalerrorHandler" ref="MULE:ERROR_HANDLER">
         <on-error-continue enableNotifications="true" logException="true" type="Attribute" when="Attribute">
             <logger/>
         </on-error-continue>
         <on-error-propagate enableNotifications="true" logException="true" type="Attribute" when="Attribute">
             <logger/>
         </on-error-propagate>
+        <logger/>
```
alegmarra: this should actually be the `on-error` element, but the test is not as powerful yet to find the correct processor for the given stereotype restriction

PR #467, opened 2018-01-12
MULE-14425: smart connector using the java module

fernandezlautaro:


extensions-xml-support/src/test/resources/flows/flows-with-module-using-java.xml
```diff
@@ -0,0 +1,32 @@
...
+    <flow name="invoke-method-flow">
+        <module-using-java:invoke-method class="java.util.HashMap" method="#[vars.method]" instance="#[vars.instance]" />
+    </flow>
+
+    <flow name="invoke-method-with-args-flow">
+        <module-using-java:invoke-method-with-args class="java.util.HashMap" method="#[vars.method]" args="#[vars.args]" instance="#[vars.instance]" />
+    </flow>
+
+    <flow name="new-method-flow">
+        <module-using-java:new-method class="java.util.HashMap" constructor="HashMap()" />
```
alegmarra: this seems redundant with `new-method-custom-class-flow`

extensions-xml-support/src/test/resources/modules/module-using-java.xml
```diff
@@ -0,0 +1,58 @@
...
+    <operation name="invoke-method">
+        <parameters>
+            <parameter name="class" type="string"/>
+            <parameter name="method" type="string"/>
+            <parameter name="instance" type="string"/>
+        </parameters>
+        <body>
+            <java:invoke class="#[vars.class]" method="#[vars.method]" instance="#[vars.instance]"/>
+        </body>
+        <output type="any"/>
```
alegmarra: will this propagate the DS of the java:invoke?

extensions-xml-support/src/test/resources/modules/module-using-java.xml
```diff
@@ -0,0 +1,58 @@
...
+    </operation>
+
+    <operation name="invoke-method-with-args">
+        <parameters>
+            <parameter name="class" type="string"/>
+            <parameter name="method" type="string"/>
+            <parameter name="instance" type="string"/>
+            <parameter name="args" type="string" use="OPTIONAL"/>
```
alegmarra: if not provided, will the `<java:args>` be left without any data so the `defaultValue="#[payload]"` kicks in, or null here means  `<java:args>#[null]</java:args>`?

extensions-xml-support/src/test/resources/flows/flows-with-module-using-java.xml
fernandezlautaro: Both test flows (with and without `...-custom-class`) are there to show not only java libraries (HashMap) are accesible, but also compiled classes within the connector (probably most common use case when using java module thru a smart connector)

extensions-xml-support/src/test/resources/modules/module-using-java.xml
fernandezlautaro: no, it will not.

extensions-xml-support/src/test/resources/modules/module-using-java.xml
fernandezlautaro: java null

PR #492, opened 2018-03-09
MULE-13354 Introduce split-aggregate router consistent with scatter-g…

dfeist:
…ather.

### DIFF
--- a/integration/src/test/java/org/mule/test/routing/SplitAggregateScopeTestCase.java
+++ b/integration/src/test/java/org/mule/test/routing/SplitAggregateScopeTestCase.java
@@ -8,46 +8,42 @@
 package org.mule.test.routing;
 
 import static java.lang.Thread.currentThread;
+import static java.util.Arrays.asList;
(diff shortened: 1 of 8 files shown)


integration/src/test/java/org/mule/test/routing/SplitAggregateScopeTestCase.java
```diff
@@ -0,0 +1,202 @@
...
+    try {
+      flowRunner(flow).run();
+      fail("Was expecting a failure");
+    } catch (Exception e) {
+      assertThat(e.getCause(), withClassName("org.mule.runtime.core.privileged.routing.CompositeRoutingException"));
+
+      Throwable compositeRoutingException = e.getCause();
+      assertThat(compositeRoutingException.getMessage(), startsWith(exceptionMessageStart));
+
+      List<org.mule.runtime.api.message.Error> exceptions = ((ComposedErrorException) compositeRoutingException).getErrors();
```
elrodro83: should do assertions on the errorType and location too.

PR #511, opened 2018-04-05
MULE-14735 - Avoid iterating parameters map on operation policies

pablolagreca:


integration/pom.xml
```diff
@@ -305,6 +314,24 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>activemq-client</artifactId>
+            <version>${activemqVersion}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.fusesource.hawtbuf</groupId>
```
estebanwasinger: why are these required?, same with the protobuf dependency

PR #514, opened 2018-04-09
MULE-14754: Flowstack: Exception when flow-ref is combined with failing

elrodro83:
interceptor and smart connector


integration/src/test/java/org/mule/test/integration/exceptions/LogCheckFlowStackTestCase.java
```diff
@@ -0,0 +1,54 @@
...
+  @Rule
+  public SystemProperty logFlowStack = new SystemProperty(MULE_FLOW_TRACE, Boolean.toString(true));
+
+  @Test
+  public void runCheckFlowStack() throws Exception {
+    runSuccesses(true, "checkFlowStack");
+  }
+
+  protected void runSuccesses(boolean verboseExceptions, String flowName) throws Exception {
+    System.setProperty(MULE_VERBOSE_EXCEPTIONS, Boolean.toString(verboseExceptions));
```
gsfernandes: Shouldn't this be set as @ Rule too?

PR #537, opened 2018-06-06
MULE-15090: Error when parsing two expressions reading a stream

afelisatti:


integration/src/test/java/org/mule/test/integration/streaming/LogRepeatableStreamTestCase.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class LogRepeatableStreamTestCase extends AbstractIntegrationTestCase {
```
marianogonzalez: this doesn't really represent what you're testing. Logging a repeatable stream works fine. The problem is when you log a message which uses several DW expressions over the same stream. Either rename the class or use allure to hint that

PR #563, opened 2018-08-07
Fixing MuleConfigurationTestCase

lucianoRM:


integration/src/test/java/org/mule/test/config/MuleConfigurationTestCase.java
```diff
@@ -36,6 +45,23 @@
   protected String workingDirectory = "target";
   private MuleContext muleContext;
 
+  private static final ClusterService mockedClusterService = mock(ClusterService.class);
```
marianogonzalez: these two shouldn't be static. Have them as instance fields and use `@Mock` with the Mockito runner

integration/src/test/java/org/mule/test/config/MuleConfigurationTestCase.java
```diff
@@ -36,6 +45,23 @@
...
+  private static final ConfigurationProperties mockedConfigurationProperties = mock(ConfigurationProperties.class);
+
+  @BeforeClass
+  public static void configureMocks() {
+    when(mockedClusterService.isPrimaryPollingInstance()).thenReturn(true);
```
marianogonzalez: this should go on the before()

integration/src/test/java/org/mule/test/config/MuleConfigurationTestCase.java
```diff
@@ -36,6 +45,23 @@
...
+    when(mockedConfigurationProperties.resolveStringProperty(anyString())).thenReturn(Optional.empty());
```
marianogonzalez: static imports

integration/src/test/java/org/mule/test/config/MuleConfigurationTestCase.java
```diff
@@ -24,18 +30,39 @@
...
 
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
 
+@RunWith(MockitoJUnitRunner.class)
```
marianogonzalez: also add `@SmallTest`

PR #571, opened 2018-08-24
MULE-15204 - Core Component definitions are created once per app

pablolagreca:


pablolagreca:
I've created the PRs using the pulverizer script. I will close the PR.

PR #576, opened 2018-08-31
MULE-15575: ArtifactDeclaration has no type information for simple pa…

pablokraan:
…rameter values

_ Updating test to include simple parameter types
_ Updating serialization test to check declarations serialized to JSON

### DIFF
--- a/dsl/src/test/java/org/mule/test/extension/dsl/ArtifactDeclarationSerializerTestCase.java
+++ b/dsl/src/test/java/org/mule/test/extension/dsl/ArtifactDeclarationSerializerTestCase.java
@@ -6,6 +6,7 @@
  */
 package org.mule.test.extension.dsl;
 
+import static java.lang.Thread.currentThread;
 import static java.util.Arrays.asList;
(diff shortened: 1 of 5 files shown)


dsl/src/test/java/org/mule/test/extension/dsl/ArtifactDeclarationSerializerTestCase.java
```diff
@@ -582,38 +618,58 @@ private static Object createNoMuleComponentsArtifactDeclaration() {
...
                 .getDeclaration())
             .withComponent(http.newOperation("request")
                 .withConfig("httpRequester")
                 .withParameterGroup(newParameterGroup("URI Settings")
-                    .withParameter("path", "/nested")
+                    .withParameter("path", createStringParameter("/nested"))
                     .getDeclaration())
                 .withParameterGroup(newParameterGroup()
-                    .withParameter("method", "POST")
+                    .withParameter("method", createStringParameter("POST"))
```
alegmarra: this is not actually necessary, right? the user can still do `.withParameter("method", "POST")`

PR #602, opened 2018-10-11
MULE-15694: Variables and Payload values set inside UntilSuccessful are not propagated outside of it in case of retry exhausted

mbuchwald:


integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulPropagationTestCase.java
```diff
@@ -0,0 +1,61 @@
...
+package org.mule.test.integration.routing.outbound;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import org.junit.Test;
+import org.mule.runtime.core.api.event.CoreEvent;
+import org.mule.test.AbstractIntegrationTestCase;
+
+public class UntilSuccessfulPropagationTestCase extends AbstractIntegrationTestCase {
```
elrodro83: add allure annotations/descriptions

integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulPropagationTestCase.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Override
+  protected String getConfigFile() {
+    return "org/mule/test/integration/routing/outbound/until-successful-propagation.xml";
+  }
+
+  @Test
+  public void variablePropagationToErrorHandler() throws Exception {
+    assertPayloadContent(flowRunner("variablePropagationToErrorHandler").withPayload("message").run());
```
elrodro83: use a parameterized test instead of repeating the test all over and just changing the flow name.

PR #616, opened 2018-11-07
MULE-15854: Error mappings should be auto created when lazy context i…

gsfernandes:
…s used

### DIFF
--- a/integration/src/test/resources/org/mule/test/integration/locator/component-locator-error-mapping.xml
+++ b/integration/src/test/resources/org/mule/test/integration/locator/component-locator-error-mapping.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


integration/src/test/resources/org/mule/test/integration/locator/component-locator-error-mapping.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:http="http://www.mulesoft.org/schema/mule/http"
+      xsi:schemaLocation="
+       http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+       http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd">
+
+    <http:listener-config name="HTTP_Listener_config">
```
afelisatti: Remove the configs, they are expensive and would require dynamic ports. You can use the `url` option for the mapping HTTP request which does not require an explicit config.

PR #629, opened 2018-12-13
MULE-16205: IdempotentRedeliveryPolicy blocks CPU_LITE thread

elrodro83:


integration/src/test/java/org/mule/test/components/RedeliveryPolicyTestCase.java
```diff
@@ -32,16 +49,54 @@ protected String getConfigFile() {
...
   }
 
+  private void sendDataWeaveObjectMessage(String flowName) throws Exception {
+    flowRunner(flowName)
+        .withPayload("{ \"name\" : \"bruce\"}")
+        .withMediaType(MediaType.APPLICATION_JSON)
+        .run();
+  }
+
+  private static CountDownLatch latch;
```
marianogonzalez: static fields and classes should be defined before instance members

integration/src/test/java/org/mule/test/components/RedeliveryPolicyTestCase.java
```diff
@@ -32,16 +49,54 @@ protected String getConfigFile() {
...
+  private static AtomicInteger awaiting = new AtomicInteger();
```
marianogonzalez: static fields should be upper case

integration/src/test/java/org/mule/test/components/RedeliveryPolicyTestCase.java
```diff
@@ -6,23 +6,40 @@
...
-  public void setUp() {
+  public void before() {
     queueHandler = new TestConnectorQueueHandler(registry);
+    latch = new CountDownLatch(1);
+    awaiting.set(0);
+  }
+
+  @After
+  public void after() throws Exception {
+    latch.countDown();
```
marianogonzalez: what's this for? looks like a left over

integration/src/test/java/org/mule/test/components/RedeliveryPolicyTestCase.java
elrodro83: only if they were `final`

integration/src/test/java/org/mule/test/components/RedeliveryPolicyTestCase.java
elrodro83: updated the test

PR #666, opened 2019-01-23
MULE-16282: workaround to work with REEs - Revert of revert of MULE-14532

mbuchwald:


mbuchwald:
Closing this to avoid hawk to fail after comments on CE

PR #671, opened 2019-01-31
MULE-16396: OnErrorPropagate should only rollback owned transactions

mbuchwald:


integration/src/test/java/org/mule/test/integration/transaction/TransactionRollbackedByOwnerTestCase.java
```diff
@@ -0,0 +1,119 @@
...
+    notifications = new ArrayList<>();
+  }
+
+  @Test
+  public void testInitializeIsCalledInInnerExceptionStrategy() throws Exception {
+    final TransactionNotificationListener listener = new TransactionNotificationListener<TransactionNotification>() {
+
+      @Override
+      public boolean isBlocking() {
+        return false;
```
marianogonzalez: consider doing this blocking to prevent race conditions when running in jenkins crappy machines

PR #685, opened 2019-03-19
MULE-16537: Error handling not working properly with foreach - Regression caused by MULE-16374

mbuchwald:


integration/src/test/resources/raise-error-config.xml
```diff
@@ -121,4 +121,21 @@
...
+                <raise-error type="APP:MY_TYPE"/>
+                <error-handler>
+                    <on-error-propagate>
+                        <logger message="Hello"/>
+                    </on-error-propagate>
+                </error-handler>
+            </try>
+        </foreach>
+        <error-handler>
+            <on-error-continue type="MULE:ANY"></on-error-continue>
```
afelisatti: Shouldn't we validate explicitly if the continue is handling the error? If we don't and the raise error stops working this would show up as OK.

PR #698, opened 2019-04-04
MULE-16411: Fix issues affecting mule-integration tests on Java 11

elrodro83:
* Support Java 11 in soap-service test support

### DIFF
--- a/soap-connect/integration-tests/src/test/java/org/mule/it/soap/connect/SoapFootballExtensionArtifactFunctionalTestCase.java
+++ b/soap-connect/integration-tests/src/test/java/org/mule/it/soap/connect/SoapFootballExtensionArtifactFunctionalTestCase.java
@@ -7,10 +7,8 @@
 package org.mule.it.soap.connect;
 
 import org.mule.functional.junit4.MuleArtifactFunctionalTestCase;
-import org.mule.it.soap.connect.services.FootballService;
-import org.mule.it.soap.connect.services.LaLigaService;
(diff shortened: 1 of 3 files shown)


soap-connect/integration-tests/src/test/java/org/mule/it/soap/connect/SoapFootballExtensionArtifactFunctionalTestCase.java
```diff
@@ -25,14 +23,22 @@
   @Rule
   public DynamicPort laLigaPort = new DynamicPort("laLigaPort");
 
-  private HttpServer footballService = new HttpServer(footballPort.getNumber(), null, null, new FootballService());
-  private HttpServer laLigaService = new HttpServer(laLigaPort.getNumber(), null, null, new LaLigaService());
+  @Rule
+  public final ExternalProcess footballServiceServer =
+      new ExternalProcess("java", "-cp", System.getProperty("soapHttpServerClasspath"), "org.mule.service.soap.server.HttpServer",
```
marianogonzalez: why can't you just start jetty programatically? Why does it have to be off process?

elrodro83: I need this to be an independent JVM so it has a plain classloader. In the future this should be an independent docker image as @juandesi suggested.

PR #706, opened 2019-05-08
MULE-16790: Global ErrorHandler does not resolve transaction

mbuchwald:


integration/src/test/java/org/mule/test/integration/exceptions/ErrorHandlerWorksWithTxTestCase.java
```diff
@@ -0,0 +1,91 @@
...
+    threads.clear();
+  }
+
+  @Test
+  public void doesNotChangeThread() throws Exception {
+    Event event = flowRunner("flowWithTx").run();
+
+    assertThat(event.getMessage().getPayload().getValue(), is("zaraza"));
+    assertThat(threads, hasSize(2));
+    assertThat(threads.get(0), is(threads.get(1)));
```
elrodro83: in a non-blocking world, the callback task may execute in the same thread. Set something in the thread local and assert against that instead of relying just in the thread identity.

integration/src/test/java/org/mule/test/integration/exceptions/ErrorHandlerWorksWithTxTestCase.java
```diff
@@ -0,0 +1,100 @@
...
+    });
+  }
+
+  @Override
+  protected String getConfigFile() {
+    return config;
+  }
+
+  public ErrorHandlerWorksWithTxTestCase(String type, String config, String errorHandlerName) {
+    System.setProperty(ERROR_HANDLER_REF, errorHandlerName);
```
elrodro83: use a `@Rule`

integration/src/test/java/org/mule/test/integration/transaction/TransactionalTryTestCase.java
```diff
@@ -11,28 +11,48 @@
...
 
   @Inject
   @Named("standardTry")
   private Flow standardTryFlow;
 
+  private String config;
+
+  @Parameterized.Parameters(name = "{0}")
+  public static Collection<Object[]> data() {
+    return Arrays.asList(new Object[][] {
```
elrodro83: in this case, you may use just an array instead of a matrix

PR #723, opened 2019-05-29
MULE-15991:No transformers for Iterator to byte[] or String cause error.

emariotti3:


integration/src/test/java/org/mule/test/integration/connection/AttemptToSendIteratorThroughHttpResponseFailTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  protected String getConfigFile() {
+    return "transformer-test-config.xml";
+  }
+
+  @Test
+  public void iteratorToByteArrayTransformerThrowsException() throws Exception {
+    HttpResponse response = Request.Post(getUrl("test/listener")).execute().returnResponse();
+    System.out.println(IOUtils.toString(response.getEntity().getContent()));
```
elrodro83: do an assert instead of a sysout

integration/src/test/java/org/mule/test/integration/connection/AttemptToSendIteratorThroughHttpResponseFailTest.java
```diff
@@ -0,0 +1,46 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.IsEqual.equalTo;
+
+public class AttemptToSendIteratorThroughHttpResponseFailTest extends LazyConnectionsTestCase {
+
+  @Rule
+  public DynamicPort port = new DynamicPort("port");
+
+  @Rule
+  public TestHttpClient httpClient = new TestHttpClient.Builder(getService(HttpService.class)).build();
```
afelisatti: Why do you need this?

PR #739, opened 2019-06-10
MULE-16981: `scatter-gather` and `choice` within transaction test

elrodro83:


integration/src/test/java/org/mule/test/integration/routing/ChoiceRouterTestCase.java
```diff
@@ -89,5 +113,51 @@ private void assertMultipleErrors(String expression, String expected) throws Exc
     }
   }
 
+  @Test
+  public void txWithNonBlockingRoute() throws Exception {
+    Message result = flowRunner("txNonBlocking").withPayload("nonBlocking").run().getMessage();
+    assertThat(capturedThreads, hasSize(1));
```
marianogonzalez: is this enough? shouldn't you also verify that this one thread is the same one that was running before entering the scope?

elrodro83: added before and after check

PR #746, opened 2019-06-14
MULE-16892: Error handler being executed twice when flow ref is involved

mbuchwald:


integration/src/test/java/org/mule/test/integration/exceptions/ErrorHandlingWithFlowRefTestCase.java
```diff
@@ -0,0 +1,115 @@
...
+import org.mule.runtime.core.api.event.CoreEvent;
+import org.mule.runtime.core.api.processor.Processor;
+import org.mule.test.AbstractIntegrationTestCase;
+
+public class ErrorHandlingWithFlowRefTestCase extends AbstractIntegrationTestCase {
+
+  private static int executions;
+
+  @Override
+  public void doSetUp() throws Exception {
```
elrodro83: use a different method annotated with `@Before` instead of overriding the before method from the superclass.

integration/src/test/java/org/mule/test/integration/exceptions/ErrorHandlingWithFlowRefTestCase.java
```diff
@@ -0,0 +1,115 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.CoreMatchers.is;
+import org.junit.Test;
+import org.mule.runtime.api.exception.MuleException;
+import org.mule.runtime.core.api.event.CoreEvent;
+import org.mule.runtime.core.api.processor.Processor;
+import org.mule.test.AbstractIntegrationTestCase;
+
+public class ErrorHandlingWithFlowRefTestCase extends AbstractIntegrationTestCase {
```
elrodro83: allure, put `@Issue` annotation also

PR #761, opened 2019-07-01
MULE-17010: OAuth uses wrong access token when multiple concurrent refresh requests are sent

elrodro83:


oauth/src/test/java/org/mule/test/module/extension/oauth/clientcredentials/OAuthClientCredentialsExtensionTestCase.java
```diff
@@ -47,7 +48,17 @@ protected void doSetUp() throws Exception {
         .withStatus(OK.getStatusCode())
         .withBody(accessTokenContent())
         .withHeader(CONTENT_TYPE, "application/json")));
+
+    // This needs to be pre-populated before the app starts, since the token will be refreshed during startup and that needs the
```
marianogonzalez: but where is it being prepopulated? You could use the object store module instead of having all this code. That would not also be simpler code but closer to the actual use case.

elrodro83: meant created, not prepopulated, will fix that

elrodro83: as for the prepopulation, i created an issue to migrate the oauth tests to MTF contemplating that

PR #782, opened 2019-07-24
MULE-17023: Nested Try is rolling back tx created by outer try

mbuchwald:


integration/src/test/java/org/mule/test/integration/transaction/TransactionsWithRoutersTestCase.java
```diff
@@ -212,17 +212,20 @@ public void flowRefToFlowWithErrorAndPropagate() throws Exception {
...
 
   @Test
-  @Ignore("MULE-17023")
   public void innerTryWithOnErrorPropagate() throws Exception {
     runsInSameThread("tryWithInnerTryWithOnErrorPropagate", TX_MESSAGE, TX_MESSAGE, OTHER_TX_MESSAGE, OTHER_TX_MESSAGE);
   }
 
+  @Test
+  public void nestedTriesWithOnErrorContinue() throws Exception {
+    runsInSameThread("nestedTriesWithOnErrorContinue", TX_MESSAGE, TX_MESSAGE, OTHER_TX_MESSAGE, OTHER_TX_MESSAGE);
```
elrodro83: is this sameThread assertion just to check there was a tx? If so, check for an active tx also

mbuchwald: Yes, the function name is a bit deprecated, since it really checks that tx is active, threads are the same and paylaods are the expected ones

PR #797, opened 2019-08-12
MULE-17313: Transaction began using TryScope are not switching to IO

mbuchwald:


integration/src/test/java/org/mule/test/integration/transaction/TransactionsWithRoutersTestCase.java
```diff
@@ -282,6 +326,10 @@ private void runsInSameTransactionWithErrors(String flow) throws Exception {
     runsInSameTransaction(flow, TX_MESSAGE, OTHER_TX_MESSAGE, "Error with " + TX_MESSAGE, OTHER_TX_MESSAGE);
   }
 
+  private void assertThreadType(Thread thread, String type) {
+    assertThat(thread.getName().matches("^\\[MuleRuntime\\]\\." + type + ".*"), is(true));
```
mbuchwald: Pattern Matcher seems to be able since hamcrest 2 (we are using 1.3)

PR #829, opened 2019-09-25
MULE-17185: Static flow-ref cycle detection test cases

plbalbi:


integration/src/test/java/org/mule/test/construct/FlowRefTestCase.java
```diff
@@ -297,6 +308,20 @@ public void backpressureFlowRefMaxConcurrencySub() throws Exception {
     assertThat(httpClient.send(request).getStatusCode(), is(SERVICE_UNAVAILABLE.getStatusCode()));
   }
 
+  private void testRecursiveFlowrefsAreDetectedFor(String callingFlowName, String offendingFlowName) {
+    try {
+      // This will attempt to start the flow. That's the moment the subscription is triggered from downstream, and that's where
+      // the inter-flow-ref cycle is checked.
+      flowRunner(callingFlowName);
+      fail();
```
elrodro83: add a message to the failure indication what might have gone wrong

elrodro83: or use an exceptionMatcher rule

plbalbi: Tried used the exception matched, but the error we are validating is thrown from the FlowRunner constructor. Will add a message.

PR #854, opened 2019-10-15
MULE-17627: fix tests

marianogonzalez:


integration/src/test/java/org/mule/test/components/RedeliveryPolicyTestCase.java
```diff
@@ -69,19 +67,6 @@ public void hashWorksOverDataWeaveObject() throws Exception {
     assertThat(queueHandler.read("redeliveredMessageQueue", RECEIVE_TIMEOUT), notNullValue());
   }
 
-  @Test
-  public void redeliveryPolicyDoesntUseCpuLite() throws Exception {
```
elrodro83: we should not lose coverage for this scenario. Parameterize it to continue executing with the proactor PS

integration/src/test/java/org/mule/test/integration/transaction/TransactionsWithRoutersTestCase.java
```diff
@@ -6,50 +6,82 @@
...
   private static List<Thread> threads;
   private static List<String> payloads;
   private static List<Boolean> runsInTx;
   private static Latch latch;
 
+  @Parameterized.Parameters
+  public static List<Object[]> parameters() {
+    return asList(
+                  new Object[] {DEFAULT_PROCESSING_STRATEGY},
+                  new Object[] {PROACTOR_PROCESSING_STRATEGY});
```
elrodro83: No need to crate the `Object[]` for single value parameterizations

integration/src/test/resources/non-blocking-test-config-global-err.xml
```diff
@@ -168,7 +168,6 @@
     </flow>
 
     <flow name="transactionalTry" initialState="stopped">
-        <test:assert-intercepting responseSameTask="false" responseExpression="#[payload == 'Test Message']"/>
```
elrodro83: we should not lose coverage for this scenario. Parameterize it to continue executing with the proactor PS

integration/src/test/resources/non-blocking-test-config.xml
```diff
@@ -156,7 +156,6 @@
     </flow>
 
     <flow name="transactionalTry" initialState="stopped">
-        <test:assert-intercepting responseSameTask="false" responseExpression="#[payload == 'Test Message']"/>
```
elrodro83: we should not lose coverage for this scenario. Parameterize it to continue executing with the proactor PS

PR #866, opened 2019-11-01
MULE-16527: Adding some Until Successful edge cases tests

plbalbi:


integration/src/test/java/org/mule/test/routing/UntilSuccessfulTestCase.java
```diff
@@ -76,6 +76,43 @@ protected void doTearDown() throws Exception {
     super.doTearDown();
   }
 
+  @Test
+  public void executesOnceWhenNoErrorArisesAndClearsContext() throws Exception {
+    CoreEvent response = flowRunner("happy-path-scope").run();
+    assertThat(getPayloadAsString(response.getMessage()), is("pig"));
+    assertThat(getNumberOfInvocationsFor("insideScope"), is(1));
+    Stack retryContextsStack = ((InternalEvent) response).getInternalParameter("RETRY_CTX");
```
elrodro83: asserting the state of the stack would better be unit testing, These integration tests should be more black-box.
Also, ideally, at this point, the stack should have been removed from the context

integration/src/test/java/org/mule/test/routing/UntilSuccessfulTestCase.java
```diff
@@ -76,6 +76,43 @@ protected void doTearDown() throws Exception {
...
+  @Test
+  public void nestedUntilSuccessfulScopesExecutionTimes() throws Exception {
+    CoreEvent response = flowRunner("nestedUntilSuccessfulScopes").run();
+    assertThat(getPayloadAsString(response.getMessage()), is("holis"));
+    // Each scope was executed desired amount of times
+    assertThat(getNumberOfInvocationsFor("outerScope"), is(2));
+    assertThat(getNumberOfInvocationsFor("innerScope"), is(6));
+    // Context are cleaned
+    Stack contextsStack = ((InternalEvent) response).getInternalParameter("RETRY_CTX");
+    assertThat(contextsStack.empty(), is(true));
```
elrodro83: same as previous comment

integration/src/test/java/org/mule/test/routing/UntilSuccessfulTestCase.java
```diff
@@ -193,6 +230,22 @@ public void untilSuccessfulInErrorHandler() throws Exception {
     assertThat(event.getMessage().getPayload().getValue(), is("hello"));
   }
 
+  @Test
+  public void concurrentEventsWithLimitedCores() throws Exception {
+    for (int i = 0; i < 3; i++) {
+      new Thread(() -> {
```
elrodro83: usa en `Executor` instead of a `Thread`

integration/src/test/resources/until-successful-test.xml
```diff
@@ -95,6 +111,29 @@
         </error-handler>
     </flow>
 
+    <flow name="concurrent-requests">
+        <until-successful maxRetries="1" millisBetweenRetries="1000">
+            <test:processor throwException="true"/>
```
elrodro83: use `raise` better

integration/src/test/resources/until-successful-test.xml
```diff
@@ -125,4 +164,19 @@
             </on-error-continue>
         </error-handler>
     </flow>
+
+    <flow name="nestedUntilSuccessfulScopes">
+        <until-successful maxRetries="1" millisBetweenRetries="10">
+            <test:invocation-counter name="outerScope"/>
```
elrodro83: i think it would be better to use the test queue component instead of this counter

integration/src/test/resources/until-successful-test.xml
```diff
@@ -125,4 +164,19 @@
...
+            <until-successful maxRetries="2" millisBetweenRetries="10">
+                <test:invocation-counter name="innerScope"/>
```
elrodro83: same as previous comment

integration/src/test/java/org/mule/test/routing/UntilSuccessfulTestCase.java
plbalbi: Working on removing it. Will move the tests validating the cleanup of the RetryContext to the scope unit tests.

integration/src/test/java/org/mule/test/routing/UntilSuccessfulTestCase.java
plbalbi: This test was experimental. Just removed it.

integration/src/test/resources/until-successful-test.xml
plbalbi: Just tried using the `test:queue` instead of the counter. It doesn't work for me, since the only methods exposed by it are a `read` and a `write` to the queue. I need just to examine the number of times the processor is executed.

PR #871, opened 2019-11-14
MULE-17747: Event context termination callbacks not called on

elrodro83:
some situations

### DIFF
--- a/integration/src/test/java/org/mule/test/core/context/notification/ConnectorMessageNotificationTestCase.java
+++ b/integration/src/test/java/org/mule/test/core/context/notification/ConnectorMessageNotificationTestCase.java
@@ -41,7 +41,7 @@ public class ConnectorMessageNotificationTestCase extends AbstractNotificationTe
     HttpRequest request =
         HttpRequest.builder().uri(String.format("http://localhost:%s/path", port.getNumber())).method(POST).build();
 
-    httpClient.send(request, HttpRequestOptions.builder().responseTimeout(RECEIVE_TIMEOUT * 1000).build());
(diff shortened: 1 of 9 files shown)


schedulers-tests/src/test/java/org/mule/test/module/scheduler/cron/SynchronousSchedulerTestCase.java
```diff
@@ -68,7 +76,10 @@ public void testSynchronous() throws InterruptedException {
   @Test
   public void testNonSynchronous() throws InterruptedException {
     assertThat(FooNonSync.latch.await(7000, MILLISECONDS), is(true));
-    assertThat(fooNonSync, hasSize(greaterThanOrEqualTo(2)));
+    probe(10000 + RECEIVE_TIMEOUT, 100, () -> {
```
fsgonz: I would set this in a single constants.

integration/src/test/java/org/mule/test/core/context/notification/ConnectorMessageNotificationTestCase.java
```diff
@@ -41,7 +41,7 @@ public void doTest() throws Exception {
     HttpRequest request =
         HttpRequest.builder().uri(String.format("http://localhost:%s/path", port.getNumber())).method(POST).build();
 
-    httpClient.send(request, TIMEOUT, false, null);
+    httpClient.send(request, HttpRequestOptions.builder().responseTimeout(RECEIVE_TIMEOUT * 1000).build());
```
fsgonz: I would set this in a single constant.

PR #892, opened 2019-12-04
MULE-16965: MuleRuntimeException when creating namespace for custom e…

gsfernandes:
…rror

### DIFF
--- a/integration/src/test/resources/org/mule/test/integration/locator/component-locator-error-mapping.xml
+++ b/integration/src/test/resources/org/mule/test/integration/locator/component-locator-error-mapping.xml
@@ -39,6 +39,12 @@
         </http:request>
     </flow>
 
+    <flow name="errorMappingFlow2">
+        <http:request method="GET" config-ref="HTTP_Request_configuration">
+            <error-mapping targetType="APP:ERROR_TYPE_5"/>
+        </http:request>
+    </flow>
+
     <http:request-config name="HTTP_Request_configuration">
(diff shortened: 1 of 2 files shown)


integration/src/test/resources/org/mule/test/integration/locator/component-locator-error-mapping.xml
```diff
@@ -27,11 +33,23 @@
...
     </flow>
 
-    <flow name="errorMappingFlowDeclaration">
+    <flow name="errorMappingFlow">
         <http:request method="GET" config-ref="HTTP_Request_configuration">
             <error-mapping targetType="APP:ERROR_TYPE_1"/>
         </http:request>
     </flow>
 
+    <http:request-config name="HTTP_Request_configuration">
```
afelisatti: Is this relevant?

gsfernandes: Yes, as when I send to lazyInit "errorMappingFlow" for the test, it has a dependency to the "HTTP_Request_configuration" therefore it will not be able to start the flow (actually the request)
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name '(inner bean)#8a0ab4c': Cannot resolve reference to bean 'HTTP_Request_configuration' while setting bean property 'parameters' with key [config-ref]; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named 'HTTP_Request_configuration' available

PR #900, opened 2019-12-14
MULE-17828: fix tests

marianogonzalez:


integration/src/test/java/org/mule/test/routing/ParallelForEachTestCase.java
```diff
@@ -142,8 +141,7 @@ public void routeWithNonMuleException() {
   @Description("An error in a route results in a CompositeRoutingException containing details of exceptions.")
   public void routeWithExpressionException() {
     assertRouteException("routeWithExpressionException", EXCEPTION_MESSAGE_TITLE_PREFIX
-        + "\t1: org.mule.runtime.core.api.expression.ExpressionRuntimeException: \"Script 'invalidExpr ' has errors: \n"
-        + "\tUnable to resolve reference of invalidExpr. at 1 : 1\" evaluating expression: \"invalidExpr\".",
+        + "\t1: org.mule.runtime.core.api.expression.ExpressionRuntimeException: Script 'invalidExpr ' has errors:",
```
elrodro83: did we lose the detailed error message from DW here?

integration/src/test/java/org/mule/test/routing/ScatterGatherRouterTestCase.java
```diff
@@ -148,8 +148,7 @@ public void routeWithNonMuleException() throws Exception {
   @Description("An error in a route results in a CompositeRoutingException containing details of exceptions.")
   public void routeWithExpressionException() throws Exception {
     assertRouteException("routeWithExpressionException", EXCEPTION_MESSAGE_TITLE_PREFIX
-        + "\t1: org.mule.runtime.core.api.expression.ExpressionRuntimeException: \"Script 'invalidExpr ' has errors: \n"
-        + "\tUnable to resolve reference of invalidExpr. at 1 : 1\" evaluating expression: \"invalidExpr\".",
+        + "\t1: org.mule.runtime.core.api.expression.ExpressionRuntimeException: Script 'invalidExpr ' has errors:",
```
elrodro83: same as above

integration/src/test/java/org/mule/test/routing/ParallelForEachTestCase.java
marianogonzalez: no,. they simply changed and the assesion fails because of the underlyining thing that DW does on hte failing part. This tets is just as good and not as dependant on how DW decides to format error messages

PR #905, opened 2019-12-19
MULE-17812: Refresh Dance not executed after TokenExpiredException

SebaElizalde:


oauth/src/test/java/org/mule/test/module/extension/oauth/authcode/OAuthExtensionTestCase.java
```diff
@@ -80,6 +80,15 @@ public void refreshToken() throws Exception {
     wireMock.verify(postRequestedFor(urlPathEqualTo("/" + TOKEN_PATH)));
   }
 
+  @Test
+  public void refreshTokenForPagedOperation() throws Exception {
+    receiveAccessTokenAndUserConnection();
+    WireMock.reset();
+    stubTokenUrl(accessTokenContent(ACCESS_TOKEN + "-refreshed"));
+    flowRunner("pagedOperation").withVariable(OWNER_ID_VARIABLE_NAME, getCustomOwnerId()).run();
```
marianogonzalez: this test is too simplistic. It should:
1) successfully get the first page
2) fail when you try to get the second
3) refresh
4) successfuly retrieve the second page

SebaElizalde: Test must remain like this until https://www.mulesoft.org/jira/browse/MULE-17905 is fixed.

PR #912, opened 2020-01-07
MULE-17894: Integration test

IvanAndresFritzler:
In order to check that the RETRY_EXHAUSTED error type is no longer replaced by a processor exception's error type, the test now throws a ConnectionException that maps to an ErrorType.

### DIFF
--- a/integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulRetryExhaustedTestCase.java
+++ b/integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulRetryExhaustedTestCase.java
@@ -7,18 +7,16 @@
 package org.mule.test.integration.routing.outbound;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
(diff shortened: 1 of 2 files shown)


integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulRetryExhaustedTestCase.java
```diff
@@ -34,13 +34,20 @@ protected String getConfigFile() {
   }
 
   @Test
-  public void onRetryExhaustedCallExceptionStrategy() throws Exception {
+  public void onRetryExhaustedCallExceptionStrategyWithRetryExhaustedError() throws Exception {
```
elrodro83: better add a new test case for this scenario, this one is already testing something else.

IvanAndresFritzler: Agree. I will make the changes :)

PR #924, opened 2020-02-03
MULE-18029: Fine grained Error resolution for the runtime core

IvanAndresFritzler:
Integration tests modifications for the retry exhausted scope in order to cover a test for MULE-17953. Error Type and Message validations.

### DIFF
--- a/integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulRetryExhaustedTestCase.java
+++ b/integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulRetryExhaustedTestCase.java
@@ -45,7 +45,13 @@ public class UntilSuccessfulRetryExhaustedTestCase extends AbstractIntegrationTe
   @Test
   public void onRetryExhaustedErrorTypeMustBeRetryExhausted() throws Exception {
     flowRunner("retryExhaustedCausedByConnectivityError").withPayload("message")
-        .run();
+            .runExpectingException(ErrorTypeMatcher.errorType(MuleErrors.RETRY_EXHAUSTED));
+  }
+
+  @Test
+  public void onRetryExhaustedLoggedMessageMustBeRetryExhausted() throws Exception {
+    flowRunner("retryExhaustedCausedByConnectivityErrorWithErrorHandler").withPayload("message")
+            .run();
   }
 
 }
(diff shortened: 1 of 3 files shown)


integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulRetryExhaustedTestCase.java
```diff
@@ -45,7 +45,7 @@ public void onRetryExhaustedCallExceptionStrategy() throws Exception {
   @Test
   public void onRetryExhaustedErrorTypeMustBeRetryExhausted() throws Exception {
     flowRunner("retryExhaustedCausedByConnectivityError").withPayload("message")
-        .runExpectingException(ErrorTypeMatcher.errorType(MuleErrors.RETRY_EXHAUSTED));
+        .run();
```
elrodro83: this doesn't seem right. Is there no failure at all now?

integration/src/test/resources/org/mule/test/integration/routing/outbound/until-successful-retry-exhausted.xml
```diff
@@ -22,11 +22,17 @@
...
-                <test:processor throwException="true" exceptionToThrow="org.mule.runtime.api.connection.ConnectionException" exceptionText="Connection refused"/>
-            </until-successful>
-        </try>
+        <until-successful maxRetries="1" millisBetweenRetries="10">
+            <test:processor throwException="true" exceptionToThrow="org.mule.runtime.api.connection.ConnectionException" exceptionText="Connection refused"/>
+        </until-successful>
+        <error-handler>
+            <test:on-error-check-log>
+                <test:check-summary>
+                    <test:summary-info key="Error type" value="MULE:RETRY_EXHAUSTED"/>
```
elrodro83: create a new test to test this, and keep the previous one as it were

integration/src/test/java/org/mule/test/integration/routing/outbound/UntilSuccessfulRetryExhaustedTestCase.java
IvanAndresFritzler: The purpose of the test is to validate the logged Message, which is done in the flow configuration (it thows an error if the expected log message is not present).

PR #941, opened 2020-03-02
MULE-18023: Revert the delete of the version in the pom as it is need…

anosenzo:
…ed to update versions with mvn versions:set


anosenzo:
--validate

PR #955, opened 2020-03-12
MULE-18178: Max Concurrency not honored in `flow-ref`

elrodro83:


integration/src/test/java/org/mule/test/construct/FlowRefTestCase.java
```diff
@@ -341,17 +350,39 @@ private void testRecursiveFlowrefsAreDetectedFor(String callingFlowName, String
     }
   }
 
+  @Test
+  @Issue("MULE-18178")
+  @Description("The maxConcurrency of a taget flow called via flow-ref is enforced")
```
mbuchwald:
```suggestion
  @Description("The maxConcurrency of a target flow called via flow-ref is enforced")
```

PR #967, opened 2020-04-03
MULE-18262: Allow to customize the default scheduler fixed frequency …

marianogonzalez:
…through a System Property


schedulers-tests/src/test/java/org/mule/test/integration/SchedulerTestCase.java
```diff
@@ -54,6 +55,13 @@ protected String getConfigFile() {
     return "org/mule/test/integration/scheduler-config.xml";
   }
 
+  @Test
+  public void test() {
+    SchedulerMessageSource source = (SchedulerMessageSource) locator.find(buildFromStringRepresentation("scheduler")).get(0);
+    FixedFrequencyScheduler conf = (FixedFrequencyScheduler) source.getConfiguration();
+    System.out.println(conf.getFrequency());
```
elrodro83, line 62: assert instead of sysout

marianogonzalez: this should actually be deleted. This was just a doodling accident

PR #968, opened 2020-04-04
MULE-18259: Smart connector fails if http protocol is not defined for…

gsfernandes:
… a config that contains another http config

### DIFF
--- a/integration/src/test/java/org/mule/test/integration/locator/LazyInitConfigurationComponentLocatorTestCase.java
+++ b/integration/src/test/java/org/mule/test/integration/locator/LazyInitConfigurationComponentLocatorTestCase.java
@@ -635,6 +635,8 @@ public class LazyInitConfigurationComponentLocatorTestCase extends AbstractInteg
   }
 
   @Test
+  @Issue("MULE-18259")
+  @Description("Default values for parameters declared in XML SDK configs were only working when XSD validations was enabled")
   public void lazyMuleContextSmartConnectorsWithConfigAndDefaultParameters() throws IllegalAccessException {
     lazyComponentInitializer.initializeComponents(componentLocation -> componentLocation.getLocation().equals("request"));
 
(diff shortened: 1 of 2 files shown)


gsfernandes:
Just the test to confirm this is not working. Basically what I see different between runtime vs design mode is related to default values that comes from configLines when parsed as those parameters that are not declared in the config from the application for XML SDK artifact are not part of the attributes (https://github.com/mulesoft/mule/blob/master/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/ComponentModelReader.java#L61) therefore the ComponentModel for the XML SDK config global element has only parameters and values for the one set in application's declaration and then when these parameters are extracted in https://github.com/mulesoft/mule/blob/master/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java#L227 they are not present neither populated as a rawValue is not present (which should be the default value in the declaration of the parameter in the XML SDK or if no default value is present the one from the ParameterModel, BTW this is present as getValue for the parameterAst but as Either.right()). 
So, in addition to this issue we should review when there is no default in the XML SDK parameter declaration and confirm this value is taken from the ParameterModel default value.

gsfernandes:
4.2.x works as there is custom logic to process parameters in MacroExpansionModule https://github.com/mulesoft/mule/blob/support/4.2.x/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java#L221

That method has been removed in favor of using the ComponentAst and CompenentParameterAst but when the rawParameters are coming from the default values from XSD (lazy has disabled this validation therefore defaults should come from ParameterModel) the ParameterAst for the missing parameters is resolved as either.right().
In addition to that, the properties extracted and resolved with literals were set again to the component therefore at that point the component parameters are fulfilled.

That could be fine as rawValue means that is already declared in the DSL so perhaps the fix is just to change the logic to retrieve the values from the getValue().getRight(). 
Just tested that and seems to be the right fix, only concern is the downcast to String.

PR #980, opened 2020-04-24
MULE-18343: Source is not notified of failure on ConnectionException in

fsgonz:
flow

### DIFF
--- a/integration/src/test/resources/org/mule/test/integration/source-connection-exception-response-handling-test-case.xml
+++ b/integration/src/test/resources/org/mule/test/integration/source-connection-exception-response-handling-test-case.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns:http="http://www.mulesoft.org/schema/mule/http"
+	xmlns="http://www.mulesoft.org/schema/mule/core"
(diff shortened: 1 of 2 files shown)


integration/src/test/resources/org/mule/test/integration/source-connection-exception-response-handling-test-case.xml
```diff
@@ -0,0 +1,25 @@
...
+                             http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd">
+	<http:listener-config name="HTTP_Listener_config">
+		<http:listener-connection host="localhost"
+			port="${httpPort}" />
+	</http:listener-config>
+
+	<flow name="http-listener-flow">
+		<http:listener config-ref="HTTP_Listener_config"
+			path="/" />
+		<script:execute engine="groovy">
```
elrodro83, line 19: why not a `raise`?

fsgonz: We need to verify that this is a ConnectionException, so raise is useless. We cannot use test:throw because ConnectionException needs a parameter

PR #1007, opened 2020-06-10
MULE-18454: Cannot serialize artifact declaration using the managed oAuth connection provider

SebaElizalde:


dsl/src/test/java/org/mule/test/extension/dsl/OcsArtifactDeclarationSerializerTestCase.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.AfterClass;
+import org.junit.runners.Parameterized;
+
+public class OcsArtifactDeclarationSerializerTestCase extends ArtifactDeclarationSerializerTestCase {
+
+  private static String ocsPropertyOldValue;
+
+  @Parameterized.Parameters(name = "{0}")
+  public static Collection<Object[]> data() {
+    ocsPropertyOldValue = System.setProperty(OCS_ENABLED, "true");
```
ndinu, line 28: Can't the utility method created on the other PR be used here?

dsl/src/test/java/org/mule/test/extension/dsl/OcsArtifactDeclarationSerializerTestCase.java
```diff
@@ -0,0 +1,58 @@
...
+                    .withParameter("connectionId",
+                                   createStringParameter("ocs:348573-495273958273-924852945/salesforce/john-sfdc-1k87kmjt"))
+                    .getDeclaration())
+                .getDeclaration())
+            .getDeclaration())
+        .getDeclaration();
+  }
+
+  @AfterClass
+  public static void cleanUp() {
```
ndinu, line 50: Use a @ClassRule SystemProperty for this.

dsl/src/test/java/org/mule/test/extension/dsl/OcsArtifactDeclarationSerializerTestCase.java
SebaElizalde: Not really because the property needs to be set not only to create the artifact but also during the test suite initialisation when extensions are discovered

ndinu: Would that be covered if we added a SystemProperty ClassRule?

SebaElizalde: Also tried it and it didn't work

dsl/src/test/java/org/mule/test/extension/dsl/OcsArtifactDeclarationSerializerTestCase.java
SebaElizalde: tried it and it didn't work :(

PR #1044, opened 2020-07-08
MULE-18041: Ignoring tests until resolution of backwards compatibilit…

IvanAndresFritzler:
…y issue


integration/src/test/java/org/mule/test/integration/exceptions/LogCheckTestCase.java
```diff
@@ -116,12 +117,14 @@ public void assertFailsIfNoException() throws Exception {
 
   @Test
   @Issue("MULE-18041")
+  @Ignore
```
elrodro83, line 120: add an issue key to ignored tests

PR #1064, opened 2020-07-21
MULE-18426: Interception API: Add a callback to source interceptor on…

JuLuzzi:
…ce all event contexts for an event are complete


integration/src/test/java/org/mule/test/integration/interception/SourceInterceptorFactoryTestCase.java
```diff
@@ -162,6 +184,79 @@ public void sourceInterceptedWithFailingProcessor() throws Exception {
     assertThat(heisenbergSourceInterceptionParameter.getParameters(), hasKey("onCapacityOverload"));
   }
 
+  @Test
+  public void sourceInterceptedAfterTerminated() throws Exception {
+    startFlow("sourceInterceptedAfterTerminated");
+
+    CountDownLatch afterCalledLatch = new CountDownLatch(1);
+
+    SourceCallbackInterceptor.afterTerminated = (componentLocation, eventContext) -> afterCalledLatch.countDown();
```
elrodro83, line 193: assert that the received eventContext is terminated (cast to BaseEventContext, call isTerminated). This applies to the other tests as well

PR #1067, opened 2020-07-23
MULE-18596: Update WSC connector version in integration tests

eze210:


mbuchwald:
We are using the bom now

PR #1087, opened 2020-07-31
MULE-18619: Component models not set for infrastructure types in AST

elrodro83:


schedulers-tests/src/test/java/org/mule/test/integration/schedule/PollScheduleNotificationTestCase.java
```diff
@@ -52,8 +57,21 @@ public void addServiceConfigurator(ServiceConfigurator serviceConfigurator) {
         // Nothing to do
       }
     });
-  }
+    builders.add(new AbstractConfigurationBuilder() {
 
+      @Override
+      protected void doConfigure(MuleContext muleContext) throws Exception {
+        ExtensionManager extensionManager;
```
marianogonzalez, line 64: this code is repeated. Already saw this in the other PR. Anyway this can be reused? Why not just adding this into the base AbstractMuleContextTestCAse

PR #1107, opened 2020-08-20
MULE-18649: Decorate outgoing streams in message collections with data flow statistics decorator

elrodro83:


integration/src/test/java/org/mule/test/streaming/PayloadStatisticsTestCase.java
```diff
@@ -0,0 +1,131 @@
...
+  public StatisticsEnabled withStatistics;
+
+  @Override
+  protected String getConfigFile() {
+    return "org/mule/streaming/payload-statistics-config.xml";
+  }
+
+  @Before
+  public void before() throws Throwable {
+    withStatistics = new StatisticsEnabled(() -> muleContext);
```
fsgonz, line 67: Maybe you can use the system property instead of the rule now after it is merged. Though we can merge this as is.

PR #1143, opened 2020-09-07
MULE-18684: Unwanted namespace changes when ConnectionException and MULE:CONNECTIVITY errors are involved

IvanAndresFritzler:


elrodro83 (changes requested):
add a test case for the reported issue

IvanAndresFritzler:
Working on a test case for https://www.mulesoft.org/jira/browse/SE-15940. Also pending to check with the SDK team if an additional test is needed in the DB connector (DbExceptionHandler) in order to fully validate https://www.mulesoft.org/jira/browse/SE-12798.

PR #1278, opened 2020-11-26
MULE-18016: Avoid using the jmh.lock file when just doing the sanity of the jmh microbenchmarks

mbuchwald:


performance/pom.xml
```diff
@@ -198,6 +198,28 @@
...
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <enablePerformanceTests>false</enablePerformanceTests>
+                                <contiperf.active>false</contiperf.active>
```
elrodro83, line 214: where are we still using contiperf?

elrodro83 (changes requested):

elrodro83 (approved):

PR #1300, opened 2020-12-04
MULE-19032: While doing test connectivity for OCS connections if the token has been expired a refresh token is not triggered

ndinu:


marianogonzalez (approved):

marianogonzalez (changes requested):
issue id needs to be mule

marianogonzalez (approved):

PR #1313, opened 2020-12-23
MULE-18936: Migrate assert-intercepting components so that they use an SDK extension

elrodro83:


integration/src/test/java/org/mule/test/core/NonBlockingFunctionalTestCase.java
```diff
@@ -14,18 +14,22 @@
...
 import java.util.Collection;
+import java.util.concurrent.CountDownLatch;
 
 import org.junit.Test;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
+import org.junit.runners.Parameterized.UseParametersRunnerFactory;
 
 @RunnerDelegateTo(Parameterized.class)
+@UseParametersRunnerFactory(FlakinessDetectorTestRunnerWithParametersFactory.class)
```
eze210, line 32: Should this be present?

eze210 (changes requested):

integration/src/test/java/org/mule/test/core/NonBlockingFunctionalTestCase.java
eze210: (FlakinessDetector)

elrodro83: absolutely not

eze210 (approved):

PR #1332, opened 2021-01-07
MULE-19122: Access Tokens are overwritten when using dynamic configs with client credentials

SebaElizalde:


oauth/src/test/resources/dynamic-client-credentials-oauth-extension-config.xml
```diff
@@ -10,6 +10,12 @@
     <os:config name="ObjectStore_Config" />
     <os:object-store name="customStore" persistent="false" config-ref="ObjectStore_Config"/>
 
+    <configuration>
```
marianogonzalez, line 13: why is expiration relevant to this example? You should be able to assert that the tokens are not overwritten even if they never expire, right?

oauth/src/test/java/org/mule/test/module/extension/oauth/clientcredentials/DynamicOAuthClientCredentialsExtensionTestCase.java
```diff
@@ -72,8 +82,39 @@ public void getSecondConnectionwithDynamicConfig() throws Exception {
...
+    wireMock.stubFor(post(urlPathMatching("/" + TOKEN_PATH))
+        .willReturn(aResponse()
+            .withStatus(OK.getStatusCode())
+            .withBody(accessTokenContent(REFRESH_TOKEN))
+            .withHeader(CONTENT_TYPE, "application/json")));
+    TestOAuthConnectionState secondConnectionState =
+        ((TestOAuthConnection) getConnectionFlow(ANOTHER_CLIENT_ID, ANOTHER_CLIENT_SECRET, ANOTHER_SCOPES).getMessage()
+            .getPayload().getValue()).getState();
+
+    new PollingProber(10000, 1000).check(new JUnitLambdaProbe(() -> {
```
marianogonzalez, line 104: why is the expiration a concern here?

oauth/src/test/java/org/mule/test/module/extension/oauth/clientcredentials/DynamicOAuthClientCredentialsExtensionTestCase.java
```diff
@@ -72,8 +82,39 @@ public void getSecondConnectionwithDynamicConfig() throws Exception {
...
+      assertThat(config.getDispose(), is(1));
+      return true;
+    }, "config was not disposed")); // Wait for the dynamic config to expire so that another connection is created, but the underlying dancer is still the same.
+    TestOAuthConnectionState thirdConnectionState =
+        ((TestOAuthConnection) getConnectionFlow(CLIENT_ID, CLIENT_SECRET, SCOPES).getMessage().getPayload().getValue())
+            .getState();
+
+    assertThat(firstConnectionState.getState().getAccessToken(), is(thirdConnectionState.getState().getAccessToken()));
```
marianogonzalez, line 112: isn't it better and more clear to assert the expected token on each state?

marianogonzalez (changes requested):

oauth/src/test/resources/dynamic-client-credentials-oauth-extension-config.xml
SebaElizalde: First request comes: a config and a connection provider are created. A dancer for the OAuth config is also generated, which stores a token with the config name as key and also feeds the token value to the connection provider instance which stores it in attributes.
Second request comes with different credentials: another config and connection provider are created. Another dancer for this other OAuth config is generated which stores a token under the SAME name (overwritten the previous one).
If a third request with the same credentials as the first comes immediately, it will produce a cache hit in the dynamic configuration provider returning the first config and its connection provider (this object is storing the token on one of its fields), so when it returns a connection it has the original token.
However if we let the config and its connection provider expire, when the third request comes, they are recreated and the connection provider requests the dancer (which does not expire) the access token. the dancer then goes to the store and retrieves the token (which due to this bug is the token generated by the second request) and gives it to the connection provider.

SebaElizalde: I can assert that tokens are not overwritten by checking the customStore directly, but because the fix changes the key used to store the tokens, if I try that then I cannot make a test that fails before the fix and passes after. This test does comply with that.

oauth/src/test/java/org/mule/test/module/extension/oauth/clientcredentials/DynamicOAuthClientCredentialsExtensionTestCase.java
SebaElizalde: Possibly, yeah.

marianogonzalez (approved):

PR #1364, opened 2021-02-02
MULE-18993: Generate an errorTypeRepository from the ArtifactAst

elrodro83:


integration/src/test/java/org/mule/test/integration/SchedulerInitialStateTestCase.java
```diff
@@ -39,29 +39,33 @@
...
+  private ConfigurationComponentLocator configurationComponentLocator;
+
   @Inject
   private TestQueueManager queueManager;
 
   @Parameterized.Parameter
   public boolean lazyInitEnabled;
 
-  @Parameterized.Parameters(name = "{index}: Running tests for {0} (validating XML [{2}]) ")
+  @Parameterized.Parameters(name = "lazyInit: {0}")
```
aimiebell, line 68: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

szaffarano (approved):

PR #1369, opened 2021-02-08
CMTS-180: SampleData resolution is not setting default values from mo…

gsfernandes:
…del to acting parameters


lucianoRM (approved):

gsfernandes:
--validate

PR #1390, opened 2021-03-02
MULE-19169: FirstSuccessful is not working inside ParallelForeach 

mbuchwald:


integration/src/test/java/org/mule/test/routing/FirstSuccessfulTestCase.java
```diff
@@ -98,4 +98,10 @@ public void insideErrorHandlerFailing() throws Exception {
     flowRunner("firstSuccessfulInErrorHandlerWithFailing").withPayload(Boolean.TRUE).run().getMessage();
   }
 
+  @Test
+  public void firstSuccessfulInsideParalleLForEach() throws Exception {
```
elrodro83, line 102: `@Issue`

integration/src/test/resources/first-successful-test.xml
```diff
@@ -126,4 +126,16 @@
             </on-error-continue>
         </error-handler>
     </flow>
+
+    <flow name="firstSuccessfulInParallelForEach">
+        <set-payload value="#[[1, 2]]"/>
+        <parallel-foreach>
+            <first-successful>
+                <route>
+                    <test:processor processingType="CPU_INTENSIVE"/>
```
elrodro83, line 135: can you use a queue here to also assert that this route is executed?

integration/src/test/resources/first-successful-test.xml
```diff
@@ -126,4 +126,16 @@
             </on-error-continue>
         </error-handler>
     </flow>
+
+    <flow name="firstSuccessfulInParallelForEach">
+        <set-payload value="#[[1, 2]]"/>
+        <parallel-foreach>
+            <first-successful>
+                <route>
```
elrodro83, line 134: is a first-successful with just one route actually valid?

elrodro83: add a TODO to validate this

elrodro83 (changes requested):

integration/src/test/resources/first-successful-test.xml
mbuchwald: It is valid, probably makes no sense but there is no place in which we say or validate differently

mbuchwald: (including xml or docs)

elrodro83 (approved):

PR #1515, opened 2021-08-02
Revert "MULE-19653: Remove uses of test:return-data (#1503)"

mbuchwald:
This reverts commit f13e7f9bd928be1ad352cd71823fe8fcd90726ce.


elrodro83 (changes requested):
do not add again this stuff we want to migrate away from!

PR #1605, opened 2021-10-28
MULE-19511: Support new SDK API Param.Stereotype annotation package

marianogonzalez:


SebaElizalde (commented):
Why is the dependency needed?

marianogonzalez:
> Why is the dependency needed? Where is it used?

It is used by the default enrichers in the ExtensionModelFactory. It is needed for the test that don't start a full Container with full classpath

SebaElizalde (approved):

PR #1608, opened 2021-11-09
Redelivery Policy: infinite loop when the redelivery is exhausted in a source configured with transactions

rbourbonnavarro:


rbourbonnavarro:
--validate

elrodro83 (approved):

PR #1674, opened 2022-01-18
MULE-19966: Replace SpringXmlConfigurationBuilder with an ArtifactAstConfigurationBuilder

elrodro83:


eze210 (changes requested):
Is this change intended to ignore some failing case when SERIALIZE_DESERIALIZE_AST_PROPERTY=true?

elrodro83:
Yes, this case tests a configuration of the ast generator that is not compatible with the serialization.

eze210 (approved):

PR #1698, opened 2022-03-22
MULE-20037: Add microbenchmark for the new dependency sorter

hyerans:


performance/src/main/java/org/mule/test/SorterLargeAppBenchmark.java
```diff
@@ -0,0 +1,162 @@
...
+  @BenchmarkMode(AverageTime)
+  public List<Object> largeAppGraphSorterGetSortedObjectsOnly() {
+    return graphSorter.getSortedObjects();
+  }
+
+  @Benchmark
+  @BenchmarkMode(AverageTime)
+  public List<Object> largeAppSpringSorterGetSortedObjectsOnly() {
+    return springSorter.getSortedObjects();
+  }
```
elrodro83, line 159: I think we would need the benchmark for the other part of the test (the adding of the objects) as well to get the proper times for the `getSortedObjects` alone

performance/src/main/java/org/mule/test/SorterSmallAppBenchmark.java
```diff
@@ -0,0 +1,162 @@
...
+      Config.class,
+      SecurityManager.class,
+      FlowConstruct.class,
+      MuleConfiguration.class,
+      Initialisable.class
+  };
+
+  @Override
+  protected String getConfigFile() {
+    return "benchmark/small-app.xml";
```
elrodro83, line 84: can this and the large one be parameters on the same benchmark class?

performance/src/main/java/org/mule/test/SorterLargeAppBenchmark.java
hyerans: @elrodro83 yes, we have three benchmarks per sorter (graphSorter and springSorter). 
1. benchmark for the adding objects(building graphs) part
2. benchmark for the get sorted objects (sorting objects from all the graphs we built) part
3. benchmark for both

Also, this version instantiates the sorter in the benchmark method instead of the set up. In this way, we don't reuse the same sorter multiple times that was created during the set up. The sorter that was instantiated in the set up will be only used when we measure this part.(2. get sorted objects (sorting objects from all the graphs we built) part)

performance/src/main/java/org/mule/test/SorterSmallAppBenchmark.java
hyerans: @elrodro83 Since we are extending the FunctionalTestCase, we can use 'getConfigFiles' instead to have two config files in one class. However, we need two muleContexts to get two different springRegistries for each app. This part was a bit tricky for me. Instead of using parameters, here is a link to a version that has only one benchmark class for both apps. Please let me know which version looks better! https://github.com/mulesoft/mule-integration-tests/commit/db9fa9bc68a47d1e8f5a2660b6e1e76dd00613be

elrodro83: how about doing a base class with the reusable stuff and have each concrete class with a different config? `getConfigFiles` won't work becuase that is inteneded to do a single app with many config files, not an app per file.

PR #1713, opened 2022-04-27
W-10781615 - SDK forward compatibility for Object Store - Support injection of the SDK API Object Store

dabreu:


integration/pom.xml
```diff
@@ -57,6 +57,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.mule.sdk</groupId>
+            <artifactId>mule-sdk-api</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
marianogonzalez, line 63: is the version necessary? don't we have dependency management on this? If DM not available here, then at least use a property

marianogonzalez (changes requested):

integration/src/test/java/org/mule/test/components/ExtensionWithObjectStoreTestCase.java
```diff
@@ -28,6 +29,9 @@ public class ExtensionWithObjectStoreTestCase extends AbstractIntegrationTestCas
   @Inject
   private ObjectStoreManager objectStoreManager;
 
+  @Inject
+  private org.mule.sdk.api.store.ObjectStoreManager sdkObjectStoreManager;
```
ndinu, line 33: For the case of the test done, the connector that injects the SDK-API ObjectStoreManager uses the same sdk-api version as the runtime. Would the injection work if the connector uses a newer or older version of the SDK-API?

dabreu: Not sure if I get your point...the injection would work on both runtime and sdk-api 4.5+

integration/pom.xml
dabreu: not necessary, my bad, I will remove it

marianogonzalez (approved):

PR #1721, opened 2022-05-16
W-11090843: Migrate HttpLoggingConfigTestCase from system test

sofiamorseletto:
Since the original system test was actually testing a change in the mule project https://github.com/mulesoft/mule/pull/6265, not the HTTP project, the test was migrated to an integration test which checks that debug logging can be configured per application

### DIFF
--- a/logging/src/test/java/org/mule/runtime/test/integration/logging/LogConfigurationTestCase.java
+++ b/logging/src/test/java/org/mule/runtime/test/integration/logging/LogConfigurationTestCase.java
@@ -7,6 +7,9 @@
 package org.mule.runtime.test.integration.logging;
 
 import static org.mule.runtime.api.util.MuleSystemProperties.MULE_SIMPLE_LOG;
+import static org.mule.tck.probe.PollingProber.probe;
+import static org.mule.test.infrastructure.FileContainsInLine.hasLine;
+import static org.mule.test.infrastructure.HasRegex.hasRegex;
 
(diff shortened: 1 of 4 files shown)


logging/src/test/java/org/mule/runtime/test/integration/logging/LogConfigurationTestCase.java
```diff
@@ -109,6 +118,43 @@ public void honorLog4jConfigFileForAppInDomain() throws Exception {
...
+  @Test
+  @Feature(INTEGRATIONS_TESTS)
+  @Story(LOGGER)
+  @Issue("W-11090843")
+  public void honorLog4jConfigFileForTwoAppsWithDifferentConfiguration() throws Exception {
+    String customLogAppName = "custom-log-app";
+    String notCustomLogAppName = "not-custom-log-app";
+
+    String customAppLog = "Custom Log App log";
+    String notCustomAppLog = "Not Custom Log App log";
```
elrodro83, line 130: use system properties for these and reference the system properties in the files

elrodro83, line 130: for clearness, make explicit in the field name that this is a debug log

elrodro83 (changes requested):

elrodro83 (approved):

elrodro83 (approved):

PR #1724, opened 2022-05-27
W-10965130: Log detailed information of exceptions that make up Compo…

hyerans:
…siteRoutingException


IvanAndresFritzler (changes requested):
Let's start breaking up this test case a bit. Create a new test case (a new class, for example "ForkJoinRoutersLogCheckTestCase") that will contain only your tests. Also we need to start using the @Parametrized annotation, so you will end up having only two actual tests, but let's discuss how to do that in a meeting.

PR #1727, opened 2022-06-01
Revert "W-10818447: Verify that ErrorHandlerFactoryBean does not gene…

marianorean:
…rate duplicate bean when defined as prototype (#1700)"

This reverts commit cdb679e2076ddef2d167f67fd40946f611c1901e.


marianorean:
--validate

PR #1728, opened 2022-06-02
W-11177824: flow ref validations fails for xml sdk components that declare inner flows when using lazy init

SebaElizalde:


extensions-xml-support/src/test/java/org/mule/test/integration/locator/LazySmartConnectorFlowRefTestCase.java
```diff
@@ -0,0 +1,87 @@
...
+
+import org.junit.Test;
+
+import io.qameta.allure.Description;
+import io.qameta.allure.Feature;
+import io.qameta.allure.Features;
+import io.qameta.allure.Issue;
+import org.junit.runners.Parameterized;
+
+@Features({@Feature(XML_SDK), @Feature(LAZY_INITIALIZATION)})
```
elrodro83, line 31: add flow-ref as well, don't remember if that is a feature or a stroy

elrodro83 (changes requested):

elrodro83 (approved):

PR #1824, opened 2022-09-27
W-11603893: Add a test for failing Global function read  in ComponentConfigurationBuilder that prevents app startup

mule-pachanady:
Support team reported a customer issue

Errors when deploying app with global functions after last release in 4.4.0-20210906

We made a fix in https://github.com/mulesoft/mule/pull/11777
This is an Integration test for the fix, to show that flow runs correctly after the fix.


mbuchwald (approved):

mule-pachanady:
--validate

mule-pachanady:
Just removed old commits where GPG key was stale or expired

PR #1869, opened 2022-11-14
W-11410770 Simple test to prove that Object Store is always created during OAuth

mule-pachanady:


mule-pachanady:
Bad PR

oauth/src/test/java/org/mule/test/module/extension/oauth/OAuthHandlerObjectStoreTestCase.java
```diff
@@ -0,0 +1,63 @@
...
+    return new String[] {"oauthHandlerWithObjectStoreFlows.xml"};
+  }
+
+  @Test
+  @Issue("W-11493901")
+  @Description("verifiy object store is always created")
+  public void verifyObjectStoreCreationDuringOAuth() throws Exception {
+
+    TestOAuthConnectionState connection = ((TestOAuthConnection) flowRunner("getConnection")
+        .run().getMessage().getPayload().getValue()).getState();
```
mule-pachanady, line 58: We really want to see if in OAuthHandler - ObjectStore os =
          objectStoreManager.getOrCreateObjectStore(storeName, ObjectStoreSettings.builder().persistent(true).build()) is called, but there is no direct way to verify that other than run thr debugger. Here we just verify that ObjectStore is created.

fsgonz (approved):

PR #1877, opened 2022-11-24
W-12117759: Ignore ExpressionLanguageFunctionsTestCase#checkCompatibleDataTypes temporarily in order to unblock PR validations

eze210:


asanguinetti (approved):
We know this test is broken because of [this change](https://github.com/mulesoft/data-weave/commit/2f67d4f61f1662b86b1ac3f308a939e1995b35ad). While the owning team works on a fix we need to be able to continue with our every day work, so we are going to be ignoring it temporarily.
The test should not be ignored for long, because it leaves a gap in coverage. If it takes longer than a day, the offending change should be reverted and the test unignored.

PR #1885, opened 2022-12-06
W-11995900: Add full set of attributes to configuration according to spec

fsgonz:


integration/src/test/java/org/mule/test/components/tracing/AbstractTracingIntegrationTestCase.java
```diff
@@ -0,0 +1,196 @@
...
+            .withExposedPorts(
+                              COLLECTOR_OTLP_GRPC_PORT,
+                              COLLECTOR_OTLP_HTTP_PORT,
+                              COLLECTOR_OTLP_GRPC_MTLS_PORT,
+                              COLLECTOR_OTLP_HTTP_MTLS_PORT,
+                              COLLECTOR_HEALTH_CHECK_PORT)
+            .waitingFor(Wait.forHttp("/").forPort(COLLECTOR_HEALTH_CHECK_PORT));
+
+    collector.start();
+    setProperty(MULE_OPEN_TELEMETRY_EXPORTER_ENABLED, TRUE.toString());
```
IvanAndresFritzler, line 135: Can we use the SystemProperty rule?

IvanAndresFritzler (changes requested):

integration/src/test/java/org/mule/test/components/tracing/AbstractTracingIntegrationTestCase.java
fsgonz: I want to set some system properties which dynamically change according to the collector and some of them have to be set depending on the different parameters. I don't think I can control that maintaining the test as parameterized.

IvanAndresFritzler (dismissed):

IvanAndresFritzler (dismissed):

fsgonz:
--validate

IvanAndresFritzler (approved):

PR #1927, opened 2023-01-03
W-12285074: Internal span attribute span.kind.override is showing up …

sofiamorseletto:
…on final exported span when it shouldn't


fsgonz (approved):

sofiamorseletto:
--validate

PR #1991, opened 2023-03-23
W-12337859: Change Span Status according to OTel semantic conventions…

sofiamorseletto:
… in HTTP Connector


sofiamorseletto:
--validate

PR #2001, opened 2023-04-10
W-12967336: Fix tests due to fixing split packages between mure-core and mule-core-components

mule-pachanady:
This is due to https://github.com/mulesoft/mule/pull/12364

PR #2004, opened 2023-04-17
W-12967336 Revert split packages renaming in mule-core-components and do changes in mule-core

mule-pachanady:
This reverts https://github.com/mulesoft/mule-integration-tests/pull/2001

Now we handle the split packages by making changes in core


mule-pachanady:
--validate

fsgonz (approved):

PR #2005, opened 2023-04-18
W-12733611: Implement the tracing of the connection obtention

IvanAndresFritzler:


tracing/src/test/java/org/mule/test/components/tracing/MuleSdkConnectionHandlingOpenTelemetryTracingTestCase.java
```diff
@@ -0,0 +1,169 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.test.components.tracing;
```
fsgonz, line 8: Shouldn't we create a test with the corresponding levels so that the connection span is only exported when the debug level is enabled? We can also create a ticket and integrate it in another PR. Probably the span should be [named](https://github.com/mulesoft/mule/blob/master/modules/tracing/mule-tracer-customization-api/src/main/java/org/mule/runtime/tracer/customization/api/InternalSpanNames.java#L26). Or the other way around.

fsgonz (changes requested):

fsgonz (dismissed):

fsgonz (approved):

sofiamorseletto (approved):

IvanAndresFritzler:
--validate

fsgonz (dismissed):

IvanAndresFritzler:
--validate

fsgonz (approved):

fsgonz (approved):

PR #2056, opened 2023-06-16
W-13093016: Refactor of resolution of the spanProcessor.

alopezbizon:


tracing/src/test/java/org/mule/test/components/tracing/AbstractOpenTelemetryTracingTestCase.java
```diff
@@ -102,10 +100,12 @@ public AbstractOpenTelemetryTracingTestCase(String exporterType, String schema,
     this.port = port;
     this.path = path;
     this.secure = secure;
+
+    configureCollector();
```
fsgonz, line 104: Instead of the constructor can we put in in an overridden doSetUpBeforeMuleContextCreation?

fsgonz (approved):

PR #2058, opened 2023-06-16
W-13097526: Fix flaky tests by replacing usages of httpbin with WireMock server

asanguinetti:
(cherry picks commits https://github.com/mulesoft/mule-integration-tests/commit/78400c136f9899b9407dd0b6ad3ed0804127d061 and https://github.com/mulesoft/mule-integration-tests/commit/6290a40e101ac186ad7d8ff21bbc6e948844a09b)

Fixes flakiness of `ProcessorInterceptorFactoryTestCase` and related ones.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,7 +110,18 @@
       <version>${muleSdkApiVersion}</version>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
+      <artifactId>wiremock</artifactId>
+      <version>1.58</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
   </dependencies>
 
   <repositories>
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -110,7 +110,18 @@
       <version>${muleSdkApiVersion}</version>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
+      <artifactId>wiremock</artifactId>
+      <version>1.58</version>
+      <scope>test</scope>
+      <exclusions>
```
elrodro83, line 118: put this in depMagmt, and the dependency on each repo actually using it

pom.xml
```diff
@@ -110,7 +110,18 @@
       <version>${muleSdkApiVersion}</version>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
```
elrodro83, line 113: format

elrodro83 (changes requested):

pom.xml
asanguinetti: that is a refactor we probably want in master before the support branch. I agree that we should do it. But wouldn't it be cleaner if we leave this one as a plain, unchanged cherry pick (collection) and then do the refactor in master and cherry pick that one?

mbuchwald (approved):

elrodro83 (approved):

PR #2126, opened 2023-08-28
W-12276645: Adapting Debug traces of the test to reflect the new implementation.

asanguinetti:


asanguinetti:
Discarded the revert of the EC impl and went for a full fix.

PR #2162, opened 2023-10-10
W-13829786: Avoid reading mule-module.properties for modules with module-info

elrodro83:
Fixing odd things in this tests that cause them to fail in a more restrictive modularized environment.

### DIFF
--- a/tracing/pom.xml
+++ b/tracing/pom.xml
@@ -215,11 +215,6 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.mule.runtime</groupId>
-            <artifactId>mule-tracer-configuration-api</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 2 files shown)


tracing/pom.xml
```diff
@@ -214,11 +214,6 @@
             <artifactId>mule-tracer-exporter-configuration-api</artifactId>
             <version>${project.version}</version>
         </dependency>
-        <dependency>
-            <groupId>org.mule.runtime</groupId>
-            <artifactId>mule-tracer-configuration-api</artifactId>
-            <version>${project.version}</version>
-        </dependency>
```
asanguinetti, line 221: What does this have to do with all the rest of the changes?

elrodro83: This was wrongly placed there and was working by chance. Fixing odd things in this tests that cause them to fail in a more restrictive modularized environment.

asanguinetti: Sorry I missed the PR description

PR #2213, opened 2023-12-14
W-14338813: Move transactionManager lifecycle/ioc tests to unit tests

mbuchwald:


integration/src/test/java/org/mule/test/spring/LookupNullObjectTestCase.java
```diff
@@ -31,26 +25,18 @@
...
   @Override
   protected String[] getConfigFiles() {
     return new String[] {};
   }
 
   @Test
-  public void noTransactionManagerThroughMuleContext() {
-    TransactionManager txManager = muleContext.getTransactionManager();
-    assertThat(txManager, is(nullValue()));
+  public void notFoundThroughRegistryLookup() {
```
eze210, line 34: This is kind of "assert 1 == 1". I understand that it could be testing something like "it doesn't throw any exception when the lookup should return null". If that's the case, please add a comment so it works as a regression test for such situations

eze210 (changes requested):

eze210 (approved):

PR #2232, opened 2024-01-04
W-14793818: Validate no compatibility plugin features are used

elrodro83:


dsl/src/test/java/org/mule/test/config/ast/ArtifactAstValidationsTestCase.java
```diff
@@ -77,10 +78,7 @@ public void astValidationsWithBaseRegistryOutsideRuntime() throws ConfigurationE
 
     List<ValidationResultItem> errors = doValidate(ast);
 
-    assertThat(errors, hasSize(1));
-
-    ValidationResultItem error = errors.get(0);
-    assertThat(error.getMessage(), equalTo("Missing Expression"));
+    assertThat(errors, hasItem(message(equalTo("Missing Expression"))));
```
rbourbonnavarro, line 81: In order to preserve the original intention of the test, shouldn't it be `contains` instead of `hasItem` in order to check there is one and only one item?

rbourbonnavarro (dismissed):

rbourbonnavarro (approved):

PR #2305, opened 2024-03-25
W-11730386: Increase timeout to fix flakiness

eze210:


asanguinetti (dismissed):

logging/src/test/java/org/mule/runtime/test/integration/logging/LoggingLibsSupportTestCase.java
```diff
@@ -27,20 +28,19 @@
...
 import io.qameta.allure.Story;
+import org.junit.Rule;
+import org.junit.Test;
+import org.slf4j.bridge.SLF4JBridgeHandler;
 
 @Features({@Feature(INTEGRATIONS_TESTS), @Feature(LOGGING)})
 @Story(LOGGING_LIBS_SUPPORT)
 public class LoggingLibsSupportTestCase extends AbstractFakeMuleServerTestCase {
 
+  private static final long PROBE_TIMEOUT = 5000;
```
akshaysonvane, line 42: Nit: Could we rename the variable to `LONG_PROBE_TIMEOUT` and add a comment explaining the reason?
Something like: Using a longer probe timeout here to account for the scheduler running the flow and for the async loggers to log the message.

eze210: done

akshaysonvane (approved):

PR #2330, opened 2024-04-18
W-15314786: Fix NPE when intercepting processor chain in router

akshaysonvane:


elrodro83 (changes requested):
Can you add a similar test case but for scopes?

akshaysonvane:
> Can you add a similar test case but for scopes?

We already have the scopes test covered [here](https://github.com/mulesoft/mule-integration-tests/blob/a0e4e354b68c67983dd656c124cd05ac4d2d0da1/integration/src/test/java/org/mule/test/integration/interception/ProcessorInterceptorFactoryChainTestCase.java#L116) that uses the `tap-phones` [scope](https://github.com/mulesoft/mule-test-artifacts/blob/d7062286cbdb441a37eacfd1a8a884ac05d48ef0/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java#L392)

elrodro83 (approved):

PR #2566, opened 2024-12-10
W-17394810: fix flaky test

rbourbonnavarro:


integration/src/test/java/org/mule/test/integration/transaction/TransactionRolledBackByOwnerTestCase.java
```diff
@@ -247,11 +249,20 @@ public String describeFailure() {
   private void assertCorrectStates() {
     if (ignoreExtraStates) {
       for (int i = 0; i < expectedStates.size(); i++) {
-        assertThat(states.get(i), is(expectedStates.get(i)));
+        assertThat("Expected " + expectedStates + " but obtained " + states, containsSequence(states, expectedStates), is(true));
```
asanguinetti, line 252: It doesn't look like this should be run inside a loop anymore.

asanguinetti (changes requested):

integration/src/test/java/org/mule/test/integration/transaction/TransactionRolledBackByOwnerTestCase.java
rbourbonnavarro: Absolutely not :piccard-facepalm:.

asanguinetti (approved):

PR #2581, opened 2024-12-13
W-17370109: Add tests for the situation where requester finds an error after committing the response payload

eze210:


eze210:
--validate

integration/src/test/java/org/mule/http/client/HttpClientResponseStreamingTestCase.java
```diff
@@ -0,0 +1,127 @@
...
+        outputStream.write(("""
+            HTTP/1.1 200 OK
+            Content-Type: text/plain; charset=iso-8859-1
+            Transfer-encoding: chunked
+
+            1
+            A
+            """).getBytes());
+        outputStream.flush();
+        outputStream.close();
```
asanguinetti, line 92: I'm a little bit rusty with sockets in Java, but aren't we forgetting to close the reading end?

integration/src/test/java/org/mule/http/client/HttpClientResponseStreamingTestCase.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Override
+    public void run() {
+      try {
+        ServerSocket passiveSocket = new ServerSocket(port);
+        Socket peerSocket = passiveSocket.accept();
+        passiveSocket.close();
+
+        DevNullReaderThread reader = new DevNullReaderThread(peerSocket);
+        reader.start();
```
asanguinetti, line 80: Does the reading have to be in parallel in order to reproduce the issue? if yes, may be add a comment explaining why?

If not, then we can simplify it.

integration/src/test/java/org/mule/http/client/HttpClientResponseStreamingTestCase.java
eze210: Yep -> I'm using a try-with-resources now

integration/src/test/java/org/mule/http/client/HttpClientResponseStreamingTestCase.java
eze210: I was using a thread because I just implemented a readUntilEOS logic, but it's true that using an extra-thread is a waste of resources... Replaced with a readUntilDoubleLineReturn so I can avoid the thread

eze210:
--validate

eze210:
--validate

PR #2634, opened 2025-02-01
W-16941297: SG timeout issue

anandkarandikar:
## Ticket

[W-16941297](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-16941297)

## Description
Attempting to add a test piggybacking on existing SG timeout test and a flow demonstrating use of https://github.com/mulesoft/mule-integration-tests/blob/b0be56518dea9b3e723f3af69e9db732dd5643d5/integration/src/test/resources/lazy-routes.xml#L16-L24

### Derby Database 
Tried using the existing derby database config, and use derby db in embedded mode. However, the derby db doesn't have a SLEEP equivalent function.

There's a potential solution where a Java method that invokes Thread.sleep() could be registered as a UDF (User Defined Function) but that requires a restart and also the classpath property modification. After a successful registration, there is a way to run the sql from java code as below:

```java
String sleepSQL = "VALUES org.example.derby.SleepExample.sleep(2000)";
stmt.executeQuery(sleepSQL);
```
where org.example.derby.SleepExample.sleep is the java method that calls Thread.sleep.

Although in theory it looks very procedural but it may not work for the integration tests because there's seems to be a restart step for an embedded db that runs during the lifetime of the tests. After numerous tries, it felt very uneasy to go down this route. 

### H2 Database
Having H2 database for a test requires adding dependency in the pom for example:

```xml
<dependency>
      <groupId>com.h2database</groupId>
      <artifactId>h2</artifactId>
      <version>2.1.214</version> <!-- Use your preferred version -->
</dependency>
```

### DB Config in the mule flow
Make sure to configure the flow with the following db config `H2_Database_Config` and use it in the `db:select`
```xml
<db:config name="H2_Database_Config" doc:name="Database Config" >
	<db:generic-connection url="jdbc:h2:~/dummyDB" driverClassName="org.h2.Driver" user="sa" />
</db:config>
```

### DB Alias for SLEEP
This is needed because like Derby DB, the H2 DB also doesn't have an inbuilt SLEEP function. 
```xml
<db:execute-ddl doc:name="Execute DDL" config-ref="H2_Database_Config">
	<db:sql ><![CDATA[DROP ALIAS IF EXISTS SLEEP]]></db:sql>
</db:execute-ddl>

<db:execute-ddl doc:name="Execute DDL" config-ref="H2_Database_Config">
	<db:sql ><![CDATA[CREATE ALIAS SLEEP FOR "java.lang.Thread.sleep"]]></db:sql>
</db:execute-ddl>
```

### Followed by the `flowRunner` invocation of the test flow that's reproduces the scenario

An oversimplified version. 
```xml
<flow name="select-with-timeout-inside-scatter-gather">
        <scatter-gather timeout="${scatterGather.timeout}">
            <route>
                <db:select config-ref="config">
                    <db:sql>select sleep(10)</db:sql>
                </db:select>
            </route>
            <route>
                <set-payload value="apple"/>
            </route>
        </scatter-gather>
    </flow>
```

### Verification step 
TBD - working on this, a rough idea is to use an `error-handler` in the flow and then execute sql 

```sql
SELECT EXECUTING_STATEMENT FROM INFORMATION_SCHEMA.SESSIONS
```
This is somewhat equivalent to running this in mysql

```sql
show full processlist;
```

### DIFF
--- a/integration/src/test/resources/routers/scatter-gather-test.xml
+++ b/integration/src/test/resources/routers/scatter-gather-test.xml
@@ -1,11 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <mule xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns:test="http://www.mulesoft.org/schema/mule/test"
+      xmlns:db="http://www.mulesoft.org/schema/mule/db"
       xmlns:test-components="http://www.mulesoft.org/schema/mule/test-components"
       xmlns:http="http://www.mulesoft.org/schema/mule/http"
       xmlns="http://www.mulesoft.org/schema/mule/core"
       xsi:schemaLocation="
                http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+               http://www.mulesoft.org/schema/mule/db http://www.mulesoft.org/schema/mule/db/current/mule-db.xsd
                http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd
                http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd
                http://www.mulesoft.org/schema/mule/test-components http://www.mulesoft.org/schema/mule/test-components/current/mule-test-components.xsd">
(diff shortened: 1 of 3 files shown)


integration/src/test/resources/routers/scatter-gather-test.xml
```diff
@@ -95,6 +97,26 @@
...
 
+    <db:config name="config">
+        <db:derby-connection database="test" subsubProtocol="memory" create="true" />
+    </db:config>
+
+    <flow name="select-with-timeout-inside-scatter-gather">
+        <scatter-gather timeout="${scatterGather.timeout}">>
+            <route>
+                <db:select config-ref="config">
+                    <db:sql>select sleep(10)</db:sql>
```
anandkarandikar, line 108: Why 10s ?
Referring to the earlier flow "timeout", it seems that the following set-payload
`<set-payload value="#[java!org::mule::tck::junit4::AbstractMuleContextTestCase::sleepFor(payload, 10000)]"/>`

will timeout. So, `select sleep(10)` is just following the suite.

PR #2685, opened 2025-05-29
W-18586120: handle `JavaVersion` values not present in older Runtimes

rbourbonnavarro:


elrodro83 (changes requested):
move this test to extensions-spring-support

